PR #2, opened 2021-11-02
TASK-50760: Add unit test for DocumentFileService

mkrout:


lbengayed (approved):

documents-services/src/test/java/org/exoplatform/documents/service/DocumentFileServiceTest.java
```diff
@@ -0,0 +1,208 @@
+/*
+ * Copyright (C) 2020 eXo Platform SAS.
```
plamarque: 2021

PR #22, opened 2021-11-17
Task-51352: Apply design for Load more button

lbengayed:


documents-webapp/src/main/webapp/skin/less/documents.less
```diff
@@ -18,6 +18,12 @@
             }
           }
         }
+        .border-box-sizing{
+          .loadMore{
```
SaraBoutej: We have already class for **border: none!important;** ==> https://github.com/Meeds-io/platform-ui/blob/develop/platform-ui-skin/src/main/webapp/skin/less/helpers.less#L84

And **color: @primaryColor!important;** ==> https://github.com/Meeds-io/platform-ui/blob/develop/platform-ui-skin/src/main/webapp/skin/less/helpers.less#L120

in our platform-ui **helpers.less** file

SaraBoutej (changes requested):

SaraBoutej (approved):
LGTM

PR #40, opened 2021-11-23
Task-51625: Avoid displaying file name as a link after applying sanitized-html directive and decode encoded file names.

mdiha:


documents-webapp/src/main/webapp/skin/less/documents.less
```diff
@@ -9,6 +9,11 @@
             border-bottom: none!important;
             .attachment {
               span {
+                a {
+                  color: @textColor!important;
+                  text-decoration-color: var(--allPagesGreyColorLighten1, #a8b3c5);
```
SaraBoutej, line 14: text-decoration-color: @greyColorLighten1;

SaraBoutej: Please use Color variable name 
text-decoration-color: @greyColorLighten1;

mdiha: Done thx

SaraBoutej (approved):
LGTM

PR #64, opened 2021-12-07
Task-50170: US_Filtre_Documents Add screens for search documents for both desktop and mobile

lbengayed:


documents-webapp/src/main/webapp/vue-app/documents/components/DocumentsMain.vue
```diff
@@ -8,7 +8,8 @@
       <div v-if="searchResult">
         <documents-header
           class="py-2" />
-        <documents-no-result-body />
+        <documents-no-result-body
+          :is-mobile="isMobile" />
```
mdiha, line 12: The isMobile prop should be passed to the child component

documents-webapp/src/main/webapp/vue-app/documents/components/header/DocumentsHeader.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div class="d-flex flex-row">
     <documents-header-left />
-    <v-spacer />
+    <v-spacer :class="showMobileFilter ? 'd-none' : ''" />
```
mdiha: You can simply use v-show directive

mdiha (approved):

PR #65, opened 2021-12-08
Task-7294: Adapt design for adding favorites in mobile.

mdiha:


SaraBoutej (approved):

documents-webapp/src/main/webapp/skin/less/documents.less
```diff
@@ -64,6 +68,13 @@
       display: none;
     }
 
+    .v-data-table__mobile-table-row > :last-child {
+      padding: 0;
+      margin: 0;
```
lbengayed: You can Add directly the vuetify class ma-0 pa-0 when it is a mobile

documents-webapp/src/main/webapp/skin/less/documents.less
```diff
@@ -79,6 +90,24 @@
           width: 28vh;
         }
       }
+
+      .uiIconVerticalDots {
+        position: relative;
```
lbengayed: You can Add directly the vuetify class position-relative

documents-webapp/src/main/webapp/skin/less/documents.less
mdiha: The element is not reachable since it is included in the v-data-table component.

SaraBoutej (approved):

PR #121, opened 2022-01-20
Task-53058: Fix UI issue in mobile.

mdiha:


documents-webapp/src/main/webapp/skin/less/documents.less
```diff
@@ -133,7 +133,7 @@
     .documents-header-right-actions {
       .inputDocumentsFilter {
         width: auto;
-        margin-top: 6px !important;
+        margin-top: 8px !important;
```
SaraBoutej: You can add vuetify helper class **mt-2** directly

mdiha: good catch thanks

SaraBoutej (approved):
LGTM

PR #136, opened 2022-01-27
Task-53490 : Adapt new design on icons and labels displayed after clicking on three dots.

ahmadmido:
Adapt new design on icons and labels displayed after clicking on three dots.


mkrout (approved):
LGTM

PR #138, opened 2022-01-30
Task-53385: Add action document details

lbengayed:


documents-webapp/src/main/webapp/vue-app/documents/components/body/actions/DownloadMenuAction.vue
```diff
@@ -1,13 +1,17 @@
...
     <v-icon
       size="13"
       class="pe-1 iconStyle">
       fas fa-download
     </v-icon>
     <span class="ps-1">{{ $t('documents.label.download') }}</span>
+    <v-divider
+      v-if="isMobile"
+      class="mt-2 f"
+      dark />
```
ahmadmido: Please use 'mt-2 dividerStyle' class to be compact with others action, no need to use dark also.

ahmadmido (approved):

PR #176, opened 2022-02-17
Task-53368: Measure download button clicks (#173)

lbengayed:
Dispatch an event listener when clicking in the Download button of document new app to track the number of times this button  is used


hbenali (dismissed):
Please add a description

mkrout (approved):

PR #180, opened 2022-02-22
Task-54561: Add feature flag on three dots of folder

lbengayed:
Add feature flag on three dots action of folder in order to not display it in Folder View


mkrout (approved):
LGTM

PR #200, opened 2022-03-02
Task-54181 : Add delete action back-end.

ahmadmido:
- Add delete action for documents.
- Undo delete action.
- Move to trash after deleting documents.
- Restore from trash action.


mkrout (approved):
LGTM

PR #205, opened 2022-03-03
Task-53872: Fix display of document or folder name with point char

lbengayed:
TASK-53872 : Fix display of document or folder name with point char


ahmadmido (approved):

mkrout (approved):
LGTM

PR #285, opened 2022-04-07
TASK-55456: Restrict document three dots menu actions depends on document visibility

hakermi:
…
Restrict document three dots menu actions depends on document visibility


mkrout (approved):
LGTM

PR #315, opened 2022-04-14
TASK-56299,56287: Implement Display pop-up Welcome to the new documents app and pop-up Quick access to recent documents

hakermi:
…

 Implement Display pop-up Welcome to the new documents app and pop-up Quick access to recent documents


mkrout (approved):
LGTM

PR #329, opened 2022-04-18
TASK-56068: Sort by date is not working in folder view

hakermi:
ISSUE: when getting folders with sort filter in parameter but without a query field or the filter field is not favourites, then no sort is applied on the jcr query to get nodes.
FIX: This PR should make sure to create an sql query with different filter items to allow getting ordered nodes depends on the sort field

### DIFF
--- a/documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/util/JCRDocumentsUtil.java
+++ b/documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/util/JCRDocumentsUtil.java
@@ -146,9 +146,7 @@ public class JCRDocumentsUtil {
                                            Session session,
                                            NodeIterator nodeIterator,
                                            Identity aclIdentity,
-                                           SpaceService spaceService,
-                                           int offset,
-                                           int limit) {
+                                           SpaceService spaceService) {
     List<AbstractNode> fileNodes = new ArrayList<>();
     while (nodeIterator.hasNext()) {
       Node node = nodeIterator.nextNode();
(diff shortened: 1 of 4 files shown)


documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/util/JCRDocumentsUtil.java
```diff
@@ -184,19 +186,20 @@ public static List<AbstractNode> toNodes(IdentityManager identityManager,
...
-    });
-    Collections.sort(fileNodes, new Comparator<AbstractNode>(){
-      public int compare(AbstractNode n1, AbstractNode n2) {
-        return Boolean.compare(n2.isFolder(),n1.isFolder());
-      }
-    });
-
-    if(limit < fileNodes.size()){
-      return fileNodes.subList(offset,limit);
+    if (filter.getSortField().equals(DocumentSortField.MODIFIED_DATE)) {
```
boubaker: For the sort and filtering, I think that it would be better to include those conditions inside the query that we will use to request backend (your first proposition was the good one I think https://github.com/exoplatform/documents/compare/677e182b29aed037f72934285acdbf55f355adc5..de05a81ca997b0e4f841b8b4958b296c3acc530b#diff-f5891a2641ecd8cc9f48b7122a7a6fd28c71b2c1a81798457ec6f6c47dcfed2cL258-L263)
Let's enhance this proposition to get a pertinent information from store instead of making the sort manually after retrieving all child nodes of a parent (which can lead to a very bad performances for a parent node having more than 1000 JCR nodes).

hakermi: PR updated

boubaker (commented):
Please add some UT

boubaker (approved):

PR #341, opened 2022-05-13
Task-57150: Bad display when user is redirected to shared folder via notif.

montassarhamadi:
Problem: Access to notif of shared file into a space redirected me to recent view and folder button was enabled.
fix: assign to the variable selectedView the value 'folder' which was initialized with  value 'timeline' this way we switch the view from recent to folder under the specific folder.

### DIFF
--- a/documents-webapp/src/main/webapp/vue-app/documents/components/DocumentsMain.vue
+++ b/documents-webapp/src/main/webapp/vue-app/documents/components/DocumentsMain.vue
@@ -502,11 +502,14 @@ export default {
         this.loading = true;
         this.previewMode = true;
         const documentPreviewId = queryParams.get('documentPreviewId');
+        this.selectedView = 'folder';
         this.showPreview(documentPreviewId);
-      } else  if (queryParams.has('folderId')) {
+      }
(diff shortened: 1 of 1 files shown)


documents-webapp/src/main/webapp/vue-app/documents/components/DocumentsMain.vue
```diff
@@ -515,10 +518,7 @@ export default {
           nodePath = nodePath.substring(0,nodePath.lastIndexOf('/'));
         }
         this.getFolderPath(nodePath);
-      } else {
-        this.parentFolderId=null;
-        this.fileName=null;
-        this.folderPath='';
+      } else if (this.selectedView === 'folder') {
         this.selectedView = 'timeline';
```
boubaker:
```suggestion
```
Shouldn't switch to timeline here IMO

montassarhamadi: I think the same as you. so I delete this line.

mkrout (approved):
LGTM

PR #342, opened 2022-05-13
Task-57189: UI bug issue on the feature Move to of new application document

wassimmhamdi:
Problem: The frame with the suggestions is too small.
Fix: fix the frame UI to make the spaces'name displayed

### DIFF
--- a/documents-webapp/src/main/webapp/vue-app/documents/components/body/table/DocumentsMoveSpaces.vue
+++ b/documents-webapp/src/main/webapp/vue-app/documents/components/body/table/DocumentsMoveSpaces.vue
@@ -53,7 +53,7 @@
           <v-list-item @click="updateSpace(item)">
             <div class="identitySuggester no-border mt-0">
               <v-chip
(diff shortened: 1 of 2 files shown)


documents-webapp/src/main/webapp/vue-app/documents/components/body/table/DocumentsMoveSpaces.vue
```diff
@@ -53,7 +53,7 @@
           <v-list-item @click="updateSpace(item)">
             <div class="identitySuggester no-border mt-0">
               <v-chip
-                class="identitySuggesterItem me-2 mt-2"
+                class="identitySuggesterItem me-2 col-8 mt-2"
```
boubaker: Why this col-8 ? IMO not needed.

wassimmhamdi: I resolved it @boubaker

boubaker (approved):

PR #352, opened 2022-05-18
Task-57150: Bad display when user is redirected to shared folder via notification

montassarhamadi:
Problem: Access to notif of shared file into a space redirected to `recent` view while `folder` button was enabled.
Fix: use `folder` view when a document preview is accessed via a link.


boubaker (approved):
Please use new title and description when merging

PR #356, opened 2022-05-20
Task-56789: Edit option is displayed for Non supported OO files

lbengayed:
Prior to this fix, Edit option is displayed for odp extension that is not supported OO files, since the mimeType of odp file includes 'presentation'.
With this fix, we will be sure to display Edit option only for '.docx,.pptx,.xlsx' extension.

### DIFF
--- a/documents-webapp/src/main/webapp/vue-app/documents/components/body/table/DocumentActionMenu.vue
+++ b/documents-webapp/src/main/webapp/vue-app/documents/components/body/table/DocumentActionMenu.vue
@@ -18,7 +18,11 @@ export default {
     file: {
       type: Object,
       default: null,
-    }
+    },
+    fileType: {
+      type: String,
+      default: null
+    },
   },
   data: () => ({
     menuExtensionApp: 'DocumentMenu',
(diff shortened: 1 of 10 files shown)


documents-webapp/src/main/webapp/vue-app/documents/components/body/table/DocumentActionMenu.vue
```diff
@@ -27,7 +31,8 @@ export default {
     mobileOnlyExtensions: ['favorite','details'],
     desktopOnlyExtensions: ['edit'],
     editExtensions: 'edit',
-    fileOnlyExtension: ['download','favorite','visibility']
+    fileOnlyExtension: ['download','favorite','visibility'],
+    fileExtensionsEditable: ['.docx','.pptx','.xlsx']
```
boubaker, line 35: Supported OO files are configured inside a service and thus should be retrieved in UI to match the exact supported list IMO. @mkrout WDYT ?

mkrout: +1, the list of supported files should not be hardcoded.

PR #359, opened 2022-05-24
TASK-57175: Transfer rules are not applied in documents app

hakermi:
Prior to this change, No transfer rules are applied on download and share documents in the documents app
This PR should make sure to check transfer rules on the download and visibility actions on documents

### DIFF
--- a/documents-webapp/src/main/webapp/vue-app/documents/components/body/table/DocumentActionMenu.vue
+++ b/documents-webapp/src/main/webapp/vue-app/documents/components/body/table/DocumentActionMenu.vue
@@ -13,6 +13,7 @@
   </v-list>
 </template>
 <script>
+
 export default {
   props: {
     file: {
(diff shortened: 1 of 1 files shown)


documents-webapp/src/main/webapp/vue-app/documents/js/DocumentFileService.js
```diff
@@ -265,4 +265,17 @@ export function undoDeleteDocument(documentId) {
       throw new Error('Error when undoing deleting document');
     }
   });
+}
+
+export function getDocumentsTransferRules() {
```
boubaker: this method has to be moved inside https://github.com/exoplatform/ecms/blob/develop/apps/portlet-documents/src/main/webapp/js/transferRulesService.js and you will be able to use it by `this.$transferRulesService...` as it's defined in https://github.com/exoplatform/ecms/blob/develop/apps/portlet-documents/src/main/webapp/vue-app/attachment/initComponents.js#L59-L63

hakermi: done in this PR: https://github.com/exoplatform/ecms/pull/1787

documents-webapp/src/main/webapp/vue-app/documents/components/body/table/DocumentActionMenu.vue
```diff
@@ -51,12 +58,22 @@ export default {
     isSymlink() {
       return this.file && this.file.sourceID;
     },
+    applyTransferRules(extension) {
```
boubaker:
```suggestion
    checkTransferRules(extension) {
```

boubaker (approved):

PR #380, opened 2022-05-30
TASK-57156 Improve documents sorting display in folder view

boubaker:
Prior to this change :
1/ Renamed document isn't displayed in top of files list (when sorting by `Last Update Time`) even when the modification date display is refreshed
2/ Folder/File grouping isn't made on Mobile View in folder view
3/ Recent documents view wasn't displayed in Mobile

This fix will :
1/ Revert https://github.com/exoplatform/documents/commit/522b2385d0cfe78851d8198e7ef730053c9716e9 to propose a new solution without affecting sort. In fact, this commit had modified the retrieved field to use a different field that was updated when renaming node. the fix was about changing the field used for sort, when renaming the file instead.
2/ Enable sorting even in Mobile so that grouping parameter is taken into consideration
3/ Delete a useless `display: none` added that affected the display of data table in Mobile view
4/ Remake default sort used for timeline view, lastupdate descending
5/ Reduce icon size for symlink in Mobile view
6/ Fix margin for first column in Mobile & Desktop for timeline & folder view
7/ Stabilize Breadcrumb usage in Folder view and Tree view
8/ Stabilize File Preview access from stream
9/ Optimize performances of rendered components
10/ Avoid duplicating event listeners when switching from a view to another


hakermi (approved):
LGTM

PR #384, opened 2022-05-30
Task-57767: Decoding documents' names with special characters in the Documents application

sofyenne:
ISSUE : When the user launches the documents application, an encoded list of document names with special characters is displayed.
FIX : Decode the document name .

### DIFF
--- a/documents-webapp/src/main/webapp/vue-app/documents/components/body/table/cells/DocumentsFileNameCell.vue
+++ b/documents-webapp/src/main/webapp/vue-app/documents/components/body/table/cells/DocumentsFileNameCell.vue
@@ -18,9 +18,9 @@
           v-if="!editNameMode"
           @click="openPreview()"
           class="document-title clickable hover-underline d-inline-flex"
-          :title="file.name">
+          :title="title">
           <div
-            v-sanitized-html="fileName"
+            v-sanitized-html="title"
             class="document-name text-truncate ms-4">
           </div>
           <div
@@ -117,6 +117,9 @@ export default {
     },
   }),
   computed: {
+    title() {
+      return decodeURI(this.file.name);
+    },
     lastUpdated() {
       return this.file && (this.file.modifiedDate || this.file.createdDate) || '';
     },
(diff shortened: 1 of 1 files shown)


documents-webapp/src/main/webapp/vue-app/documents/components/body/table/cells/DocumentsFileNameCell.vue
```diff
@@ -117,6 +117,9 @@ export default {
     },
   }),
   computed: {
+    title(){
+     this.file.name=decodeURI(this.file.name);
+    },
```
boubaker:
```suggestion
    title() {
      return decodeURI(this.file.name);
    },
```

sofyenne: Done !

boubaker (dismissed):

documents-webapp/src/main/webapp/vue-app/documents/components/body/table/cells/DocumentsFileNameCell.vue
```diff
@@ -238,6 +241,7 @@ export default {
     openPreview() {
       this.loading = true;
       if (this.file && this.file.folder){
+        this.file.name = decodeURI(this.file.name);
```
boubaker: we should avoid changing retrieved information and keep it as it is. BTW, you can add a dedicated data variable with different name such as `fileName` and assing into it the decoded value.

boubaker:
@sofyenne I can't reproduce the problem anymore, can you check again please

documents-webapp/src/main/webapp/vue-app/documents/components/body/table/cells/DocumentsFileNameCell.vue
sofyenne: i leave this change i find is not necessary after a test of fixed behavior

boubaker (approved):

PR #398, opened 2022-06-02
Task-57136: Translate terms when creating / duplicating files

lbengayed:
Add i18n prifix clone in order to translate terms when duplicating files.
Add unit test
Decrease the complexity of methode duplicateItem


mkrout (approved):
LGTM

PR #399, opened 2022-06-02
Task-57552: Decoding the document name in the favorite list

sofyenne:
ISSUE: If user upload file with a ( ' ) character in the documents app of a space and add the document to favorites , The document name is encoded in the favorite list.
FIX : Decode the document Title .

### DIFF
--- a/documents-webapp/src/main/webapp/vue-app/documents-extensions/components/DocumentsFavoriteItem.vue
+++ b/documents-webapp/src/main/webapp/vue-app/documents-extensions/components/DocumentsFavoriteItem.vue
@@ -47,7 +47,7 @@ export default {
     this.$attachmentService.getAttachmentById(this.id)
       .then(file => { 
         this.file = file;
-        this.documentTitle =decodeURI(decodeURI((file.title))) ;
+        this.documentTitle = decodeURI(decodeURI(file.title)) ;
         const updaterFullName = file && file.updater && file.updater.profile && file.updater.profile.fullname || '';
         const updateDate = new Date(file.updated);
         const updateDateInfo = this.$dateUtil.formatDateObjectToDisplay(updateDate, this.dateFormat);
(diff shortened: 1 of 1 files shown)


documents-webapp/src/main/webapp/vue-app/documents-extensions/components/DocumentsFavoriteItem.vue
```diff
@@ -46,6 +46,7 @@ export default {
   created() {
     this.$attachmentService.getAttachmentById(this.id)
       .then(file => { 
+        file.title = decodeURI(file.title); 
         this.file = file;
         this.documentTitle = decodeURI(file.title);
```
ahamdi: there is already a decoding of property documentTitle which takes the value of file.title. which one is used in our case ? do we need both of properties ?

sofyenne: One call of the decodeURI method won't solve the problem, i think the file title is encoded twice .That's why I decode the title and the document title. I changed this solution with a double call of decodeURI on the documentTitle.

documents-webapp/src/main/webapp/vue-app/documents-extensions/components/DocumentsFavoriteItem.vue
```diff
@@ -47,7 +47,7 @@ export default {
     this.$attachmentService.getAttachmentById(this.id)
       .then(file => { 
         this.file = file;
-        this.documentTitle = decodeURI(file.title);
+        this.documentTitle =decodeURI(decodeURI((file.title))) ;
```
ahamdi:
```suggestion
        this.documentTitle = decodeURI(decodeURI(file.title)) ;
```

ahamdi (approved):

PR #404, opened 2022-06-07
Task-54482: Add missing listener to update activity cache when attachments renamed

sofyenne:
ISSUE : When user post a short message on space using existing uploads and rename the original file , the title of document not updated in activity stream.
FIX : add missing listener to clear cache activity related to document updated.

### DIFF
--- a/documents-services/src/test/java/org/exoplatform/documents/service/DocumentFileServiceTest.java
+++ b/documents-services/src/test/java/org/exoplatform/documents/service/DocumentFileServiceTest.java
@@ -13,33 +13,58 @@
  *
  * You should have received a copy of the GNU General Public License
  * along with this program; if not, see<http://www.gnu.org/licenses/>.
-*/
+ */
 package org.exoplatform.documents.service;
 
 import static org.junit.Assert.*;
 import static org.mockito.Matchers.eq;
 import static org.mockito.Mockito.*;
 
+import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import org.exoplatform.commons.cache.future.FutureExoCache;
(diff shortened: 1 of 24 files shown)


ahamdi (changes requested):
Could you cover this change with a dedicated unit test ?

documents-services/src/test/java/org/exoplatform/documents/service/DocumentFileServiceTest.java
```diff
@@ -13,51 +13,82 @@
  *
  * You should have received a copy of the GNU General Public License
  * along with this program; if not, see<http://www.gnu.org/licenses/>.
-*/
```
ahamdi, line 16: Could you revert formatting this class, it makes validation harder

documents-services/src/test/java/org/exoplatform/documents/service/DocumentFileServiceTest.java
```diff
@@ -66,7 +97,28 @@ public void setUp() {
...
+  }
+
+  @Test
+  public void NotClearCacheActivityWhenNotcachedActivityStorage() throws Exception {
+    String documentID = "2";
+    Event<String, String> renameFileEvent = mock(Event.class);
+    cachedActivityStorage = mock(CachedActivityStorage.class);
+    AttachementsActivityCacheUpdater attachementsActivityCacheUpdater = new AttachementsActivityCacheUpdater(activityStorage);
+    attachementsActivityCacheUpdater.onEvent(renameFileEvent);
+    verify(cachedActivityStorage, times(0)).clearActivityCachedByAttachmentId(documentID);
```
ahamdi: We need to make sure that the event is called once

ahamdi (changes requested):

ahamdi (approved):
LGTM

PR #412, opened 2022-06-21
TASK-57530 fix breadcrumb related to a symlink folder

Jihed525:
Problem: when clicking on symlink folder, the BreadCrumb item which contains the name folder is replaced with the path of the original folder.on technical side, the problem was in the implementation inside function getBrBreadCrumb() in JCRDocumentFileStorage class.if the type of the node was Symlink ,the breadCrumb of source node was returned to be displayed in front.
Fix: remove the part of code which check on Symlink node to avoid replacing it with original node, then return the correct BreadCrumb Items of current folder.
(cherry-picked from https://github.com/exoplatform/documents/commit/efbbdc09e89c47dd9d05abde2d00b0070834b09c )


Jihed525:
as backport already made by @wassimmhamdi in pr https://github.com/exoplatform/documents/pull/411 i close this pr

PR #416, opened 2022-06-30
Task-58643: Fix rename document uploaded

sofyenne:
ISSUE: When user rename document uploaded the extension of document not displayed 
FIX : change the clean string method to correctly clean the new name

### DIFF
--- a/documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorage.java
+++ b/documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorage.java
@@ -536,7 +536,7 @@ public class JCRDocumentFileStorage implements DocumentFileStorage {
       String oldName = node.getName();
       if (oldName.indexOf('.') != -1 && node.isNodeType(NodeTypeConstants.NT_FILE)) {
         String ext = oldName.substring(oldName.lastIndexOf('.'));
-        title = title.concat(ext);
+        title = name.concat(ext);
         name = name.concat(ext);
       }
 
--- a/documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorage.java
+++ b/documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorage.java
@@ -529,7 +529,6 @@ public class JCRDocumentFileStorage implements DocumentFileStorage {
       }
       String name = Text.escapeIllegalJcrChars(cleanString(title));
       //clean node name
-      name = cleanString(name);
 
       name = URLDecoder.decode(name,"UTF-8");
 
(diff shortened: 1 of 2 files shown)


documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/util/JCRDocumentsUtil.java
```diff
@@ -534,7 +534,7 @@ public static String cleanString(String str) {
     int i = 0;
     while (i < strLength) {
       char c = cleanedStr.charAt(i);
-      if (c == '/' || c == ':' || c == '[' || c == ']' || c == '*' || c == '\'' || c == '"' || c == '|' || c == 'ʿ' || c == 'ˇ') {
+      if (c == '/' || c == ':' || c == '[' || c == ']' || c == '*' || c == '\'' || c == '"' || c == '|' || c == 'ʿ' || c == 'ˇ' || c == '.') {
         cleanedStr.deleteCharAt(i);
         cleanedStr.insert(i, '_');
       } else if (!(Character.isLetterOrDigit(c) || Character.isWhitespace(c) || c == '.' || c == '-' || c == '_')) {
```
ahamdi: the dot character is already removed here

ahamdi (changes requested):

documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/util/JCRDocumentsUtil.java
sofyenne: done !! i change the else statement , when rename document the string to clean is the title of document without the extension so we need to replace the dot character by "_" not remove it to keep the meaning of the name  .
example : document name = `test.test.test.pdf` => expected name is    `test_test_test.pdf `

documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorage.java
```diff
@@ -536,7 +536,7 @@ public void renameDocument(long ownerId, String documentID, String title, Identi
       String oldName = node.getName();
       if (oldName.indexOf('.') != -1 && node.isNodeType(NodeTypeConstants.NT_FILE)) {
         String ext = oldName.substring(oldName.lastIndexOf('.'));
-        title = title.concat(ext);
+        title = name.concat(ext);
```
ahamdi, line 538: No need to change the title, the title is a property that does not need to be cleaned nor escaped its illegal JCR characters

ahamdi (changes requested):
Please add a UT if possible

documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorage.java
ahamdi: Also I don't get why do we need to do cleanString twice in lines 530 and 532

sofyenne: i remove the unnecessary clean string

ahamdi (approved):

PR #430, opened 2022-08-01
Task-58997: Add the ability to create folder with the name contains only numbers

sofyenne:
ISSUE : On new document application users can't create a folder  with the name contains only numbers
FIX : Remove the condition who verify if the name of folder is contain only numbers and can't able user to create this folder


ahamdi (approved):
I do not find this restriction logical, OK for this behavior change. 
LGTM

PR #431, opened 2022-08-01
Task-58927 : Fix copied link of shared document does not work

sofyenne:
ISSUE : When users copy the link of the shared document and past it to preview this document , the document not opened and users redirect to the folder view of application .
FIX :  add `sourceId` of shared document to copied link


ahamdi (changes requested):
The preview should work for the symlink , when we get the ID of the symlink, we have to load the original content and preview it correctly

documents-webapp/src/main/webapp/vue-app/documents/components/body/table/cells/DocumentsFileNameCell.vue
```diff
@@ -243,10 +243,7 @@ export default {
       if (this.file && this.file.folder){
         this.$root.$emit('document-open-folder', this.file);
       } else {
-        let id = this.file.id;
-        if (this.file.sourceID){
-          id = this.file.sourceID;
-        }
+        let id = this.file.sourceID ? this.file.sourceID :this.file.id;
```
sofyenne, line 255: @ahamdi  the `sourceId` is the id of original file so the condition to check if the file is a symlink is already exist ( if sourceId not null so the file is a symlink ) and get the original attachment by the id or the sourceId  that's why the open preview document works correctly , i just clean the code .

documents-webapp/src/main/webapp/vue-app/documents/components/body/table/cells/DocumentsFileNameCell.vue
```diff
@@ -271,7 +268,7 @@ export default {
           })
           .catch(e => console.error(e))
           .finally(() => {
-            window.history.pushState('', '', `${eXo.env.server.portalBaseURL}?documentPreviewId=${this.file.id}`);
+            window.history.pushState('', '', `${eXo.env.server.portalBaseURL}?documentPreviewId=${this.file.sourceID ? this.file.sourceID : this.file.id}`);
```
sofyenne, line 281: @ahamdi  i add the condition to check if the file is a symlink , to get the original attachment when user reload the page

documents-webapp/src/main/webapp/vue-app/documents/components/body/actions/CopyLinkMenuAction.vue
```diff
@@ -32,7 +32,7 @@ export default {
       if (this.file.folder){
         path = `${path}?folderId=${this.file.id}`;
       } else {
-        path = `${path}?documentPreviewId=${this.file.id}`;
+        path = `${path}?documentPreviewId=${this.file.sourceID ? this.file.sourceID:this.file.id}`;
```
sofyenne, line 35: @ahamdi i add the same condition to the copied link to get the original file and that fix the problem .WDYT ?

documents-webapp/src/main/webapp/vue-app/documents/components/body/table/cells/DocumentsFileNameCell.vue
ahamdi: I agree, and think this is how it was implemented. But the behavior is not correct, we need to be able to preview a symlink the same way we preview original document.

ahamdi (changes requested):

PR #449, opened 2022-08-11
Task-59174: Display cloud drive folder in Drives app

IlhemEssaadi:
Prior to this change, cloud drive folder in Drives app is not well displayed, this fix will correct the cloud drive folder icon as mentioned in the design.

### DIFF
--- a/documents-webapp/src/main/webapp/vue-app/documents/components/body/table/cells/DocumentsFileNameCell.vue
+++ b/documents-webapp/src/main/webapp/vue-app/documents/components/body/table/cells/DocumentsFileNameCell.vue
@@ -20,8 +20,9 @@
           @click="openPreview()"
           class="document-title clickable hover-underline d-inline-flex"
           :title="title">
-          <div class="document-name text-truncate ms-4">
-          <span> {{ title }} </span>
+          <div
+            v-sanitized-html="title"
+            class="document-name text-truncate ms-4">
           </div>
           <div
             v-sanitized-html="fileType"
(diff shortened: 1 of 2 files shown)


documents-webapp/src/main/webapp/vue-app/documents/components/body/table/cells/DocumentsFileNameCell.vue
```diff
@@ -8,20 +8,20 @@
...
         <div
           v-if="!editNameMode"
           @click="openPreview()"
           class="document-title clickable hover-underline d-inline-flex"
           :title="title">
-          <div
-            v-sanitized-html="title"
-            class="document-name text-truncate ms-4">
+          <div class="document-name text-truncate ms-4">
+          <span> {{ title }} </span>
```
ahamdi: Could you explain the reason to disable title sanitization ?

ahamdi (changes requested):

documents-webapp/src/main/webapp/vue-app/documents/components/body/table/cells/DocumentsFileNameCell.vue
IlhemEssaadi: > disable title sanitization

I disabled title sanitization because the folder title contains an email address and the email should not be clickable and open a new tab to send new mail.

ahamdi: Sanitization is a must -have to protect user data, we need to find a way to disable sending emails without disabling sanitiation . Could you suggest another solution for the current issue ?

ahamdi (approved):

PR #466, opened 2022-09-05
Task-59419_59420: Customize 3 dots options on One Drive and Google Drive cloud drive folders

IlhemEssaadi:
Prior to this change, all options are allowed from 3 dots for One Drive and Google Drive cloud drive folders. After this change, we keep only copy link as option for these folders in order to avoid wrong operations.


azayati (approved):
LGTM

PR #467, opened 2022-09-06
Task-59422_59423: Don't display modify option from the 3 dots options of One Drive and Google Drive cloud drive documents

abenmbarek:
Prior to this change, modify option is allowed from 3 dots for One Drive and Google Drive cloud drive documents. After this change, this option should no more be displayed in order to avoid modifying cloud drive documents with only office.


azayati (approved):
LGTM

PR #469, opened 2022-09-07
TASK-59433: Fix rename file with accentuated characters

hchaabene:
Prior to change there is a regression caused by https://github.com/exoplatform/documents/pull/416
So the modification is reverted, no need to change the title since it is a property that does not need to be cleaned nor escaped its illegal JCR characters.


documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorage.java
```diff
@@ -538,7 +538,7 @@ public void renameDocument(long ownerId, String documentID, String title, Identi
       } else {
         node = getNodeByIdentifier(session, documentID);
       }
-      String name = Text.escapeIllegalJcrChars(cleanString(title));
```
ahamdi: we should not remove cleanString here, instead, we need just to clean the content name and preserve the title as is.

ahamdi (changes requested):

azayati (approved):
LGTM

ahamdi (approved):

PR #472, opened 2022-09-09
Task-58900: In move drawer when changing the destination folder the current location shouldn't be changed

Jihed525:
Problem: when click on the the move button of a folder in documents application of a space, the hierarchy icon of the Current Location and destination in the opened move drawer are clickable Fix: we add a property to assure that these two tree views are hidden and not clickable (cherry picked from commit 8333dff2dd0391b538e933051dce5a52e9167a4d)


azayati (approved):
LGTM

PR #482, opened 2022-09-15
TASK-59433: Rename file : accentuated charachters aren't saved

hchaabene:
Prior to change there is a regression caused by #416 So the modification is reverted, no need to change the title since it is a property that does not need to be cleaned nor escaped its illegal JCR characters.


azayati (approved):
LGTM

sofyenne:
> Prior to change there is a regression caused by #416 So the modification is reverted, no need to change the title since it is a property that does not need to be cleaned nor escaped its illegal JCR characters.

PR #487, opened 2022-09-16
Task-59789: Fix feature flag bug

abenmbarek:


azayati (changes requested):
you need to test this change in Acceptance by verifying that poll button appears again and your exo.feature.AddShortcutAction.enabled works correctly cc @mkrout

mkrout (approved):
LGTM

PR #490, opened 2022-09-19
feat: Add version history option - EXO-59667

hakermi:
Add version history option to display list of versions of a document


mkrout (approved):
lgtm

PR #492, opened 2022-09-19
fix: Fix translation of task folder name to the selected language - EXO-46507

sofyenne:
Prior this change the task folder name not translated  to the selected language of platform, this pr allow the document folder task name to be translated to the same language of platform.


azayati (changes requested):
can you update the PR title with the new format ?

sofyenne:
Done !

azayati (changes requested):
The problem is not only with task folder but also with news and notes ones.
I think that this isn't something to be fixed in documents app because even with this fix the problem should remain from backoffice /portal/g/:platform:web-contributors/siteExplorer with old documents app the problem will remain. this needs a functional/technical feedback

PR #498, opened 2022-09-21
fix: fix problem of delete document - EXO-59821

IlhemEssaadi:


azayati (changes requested):
This should be fix in the title not feat

mkrout (approved):
LGTM

mkrout (approved):
LGTM

PR #506, opened 2022-09-26
feat: Add Drag and drop - EXO-59516 (#497)

mkrout:


IlhemEssaadi (dismissed):

documents-webapp/src/main/webapp/vue-app/documents/components/body/UploadOverlay.vue
```diff
@@ -0,0 +1,55 @@
+<template>
```
abenmbarek, line 18: Licence header please

abenmbarek (approved):

PR #511, opened 2022-09-27
feat: Space members aren't notified when they are designed as collaborators on a document - EXO-59955

IlhemEssaadi:


mkrout (approved):
LGTM

PR #520, opened 2022-09-29
Task-60009: Fix refresh after delete related to EXO-57172

Jihed525:
Prior to this fix, when a user deletes a file in the documents app, the delete operation is delayed by 5 seconds to allow the user to cancel the edit if needed, so if the user refreshes the UI before the delay is over, the list of documents is retrieved before de delete is done, and the deleted file will be displayed in the list, another refresh after the delay will return the list without the file.
The fix will simply check the list of files to be deleted and don't add them to the list.
 Cherry-picked from [PR](https://github.com/exoplatform/documents/pull/516)


azayati (approved):
LGTM

PR #543, opened 2022-10-10
fix: Keep documents search term filter when switching between documents views - EXO-60059

akhanfir:
Prior to this change, when we are on a specific document view, make a document search based on a search term filter then switch to another document view, the search results are still displayed but the corresponding search term filter is removed. After this change, we keep the search term filter displayed like the search results when switching between views.


azayati (approved):
LGTM

PR #552, opened 2022-10-11
fix: Fix uploading files inside subfolders of an uppercase parent folder name - EXO-59861

Jihed525:
Prior to this change, When upload a a file inside a subfolder starts from level 2 in the hierarchy and while the parent folder has an uppercase name, the upload fails and a duplicated folder is created with the attachments drawer with lowercase name.
This PR should make sure send the right selected path.

(cherry picked from commit 4170ebedd00e48d6a22fd602de4ee9c5cbb490da)


hbenali:
Stale PR

PR #577, opened 2022-10-20
fix: Fix lost extension after renaming a document - EXO-60015 

sofyenne:
Prior this change, when we upload a file on eXo 6.3.0 and rename it, the extension is lost even if we migrate to last eXo 6.3.x or 6.4.x versions. This is caused by the clean name method which replaces the `. ` character by `-`. After this commit:
- For the first case, we check the files uploaded and not renamed on eXo 6.3.0 and recover the extension from the property exo:title when renaming this file. 
- For the second case, we check the files uploaded and renamed on eXo 6.3.0 and recover the extension, based on the file mimetype and an NtFilesExtension.json file which contains the associations between well known mimeTypes and extensions.


azayati (approved):
LGTM

PR #582, opened 2022-10-26
fix: Keep documents search term filter when switching between documents views - EXO-60059

Jihed525:
Prior to this change, when we are on a specific document view, make a document search based on a search term filter then switch to another document view, the search results are still displayed but the corresponding search term filter is removed. After this change, we keep the search term filter displayed like the search results when switching between views.

(cherry picked from commit fdaabf656459d13a88e4cbe5180d343bb1b7c60b)


azayati (approved):
LGTM

PR #601, opened 2022-11-21
fix: fix load more button missing - EXO-60826

GouadriaHanen:
before this change, load more is not displayed because the size of the folders returned is less than or equal to the limit because when filtering the folders, some of them were deleted ;
after this change, retrieved documents from the database are filtred and with offset and limit


documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorage.java
```diff
@@ -320,7 +320,7 @@ public List<AbstractNode> getFolderChildNodes(DocumentFolderFilter filter,
           String statement = getFolderDocumentsQuery(parent.getPath(), sortField, sortDirection);
           Query jcrQuery = session.getWorkspace().getQueryManager().createQuery(statement, Query.SQL);
           ((QueryImpl)jcrQuery).setOffset(offset);
```
ahamdi, line 318: shouldn't we update the offset with the index of the last returned element ? 
What if limit*2 elements got filtered and we get less than limit elements ?

ahamdi (changes requested):

ahamdi (commented):
Could you add UT for the changes ?

ahamdi (approved):

PR #603, opened 2022-11-23
fix: Fix personal drive shared document preview issues (breadcrumb, wrong folder displayed) - EXO-60838

GouadriaHanen:
prior to this change, when sending a notification for the shared document or selecting a document to display it send the UUID of the original document which resulted in an incorrect display of the breadcrumb trail and folders
prior to this change, when sharing or selecting a document it's retrieved by its UUID

### DIFF
--- a/documents-webapp/src/main/webapp/vue-app/documents/components/body/table/cells/DocumentsFileNameCell.vue
+++ b/documents-webapp/src/main/webapp/vue-app/documents/components/body/table/cells/DocumentsFileNameCell.vue
@@ -254,10 +254,7 @@ export default {
       if (this.file && this.file.folder){
         this.$root.$emit('document-open-folder', this.file);
       } else {
-        let id = this.file.id;
-        if (this.file.sourceID){
-          id = this.file.sourceID;
-        }
+        const id = this.file.id;
         this.$attachmentService.getAttachmentById(id)
           .then(attachment => {
             documentPreview.init({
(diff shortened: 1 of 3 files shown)


ahamdi (approved):

documents-webapp/src/main/webapp/vue-app/documents/components/body/table/cells/DocumentsFileNameCell.vue
```diff
@@ -254,10 +254,7 @@ export default {
       if (this.file && this.file.folder){
         this.$root.$emit('document-open-folder', this.file);
       } else {
-        let id = this.file.id;
-        if (this.file.sourceID){
-          id = this.file.sourceID;
-        }
+        const id = this.file.id;
```
azayati, line 257: check this [commit](https://github.com/exoplatform/documents/commit/6ba66ad99760d79cd6062bd1fc6108e493def269#diff-bbbe756cfeab1ec5fcb9eda9377fe51993bce6cce0d0a810072129c77d505c69L264), we should verify if that this modification will not cause a regression

azayati (dismissed):

documents-webapp/src/main/webapp/vue-app/documents/components/body/table/cells/DocumentsFileNameCell.vue
GouadriaHanen: duplicating a shared document works perfectly

PR #606, opened 2022-11-29
fix : Document uploaded in wrong location - EXO-60910

GouadriaHanen:
Before this change, after creating a document in a specific folder and opening it in the docs preview, the window path name takes the wrong path, so when creating a new document, its location depends on the window path name resulting a wrong location after upload.
After this change, when opening a document in the preview, the window path takes the correct folder location.


ahamdi (approved):
LGTM

PR #609, opened 2022-11-30
fix: Fix personal drive shared document preview issues (breadcrumb, wrong folder displayed) - EXO-60838 (#605)

Jihed525:
prior to this change, when sending a notification for the shared document or selecting a document to display it send the UUID of the original document which resulted in an incorrect display of the breadcrumb trail and folders prior to this change, when sharing or selecting a document it's retrieved by its UUID

(cherry picked from commit b3b6507903de3d61d81031044403a1131c387eaa)


documents-services/src/main/java/org/exoplatform/documents/listener/ShareDocumentNotificationListener.java
```diff
@@ -47,8 +48,17 @@ public void onEvent(Event<Identity, Node> event) throws Exception {
     Identity targetIdentity = event.getSource();
     String currentUser = ConversationState.getCurrent().getIdentity().getUserId();
     NotificationContext ctx = NotificationContextImpl.cloneInstance();
+<<<<<<< HEAD
```
GouadriaHanen: could you please resolve the conflict?

GouadriaHanen (changes requested):

GouadriaHanen (approved):

PR #612, opened 2022-12-01
fix: fix No version in the history version drawer if a file is duplicated : EXO-61044 

GouadriaHanen:
prior to this change, after duplicating a document, the history version drawer is empty since no version is added
after this change, an auto version is added after creating a document

### DIFF
--- a/documents-storage-jcr/src/test/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorageTest.java
+++ b/documents-storage-jcr/src/test/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorageTest.java
@@ -190,7 +190,6 @@ public class JCRDocumentFileStorageTest {
     when(identity.getRemoteId()).thenReturn("username");
     when(JCRDocumentsUtil.getUserSessionProvider(repositoryService, userID)).thenReturn(sessionProvider);
     when(CommonsUtils.getService(AutoVersionService.class)).thenReturn(autoVersionService);
-    when(autoVersionService.autoVersion(any(Node.class))).thenReturn(null);
     jcrDocumentFileStorage.duplicateDocument(1L,"1","copy of",userID);
     verify(sessionProvider, times(1)).close();
     verify(autoVersionService, times(1)).autoVersion(any(Node.class));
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -67,7 +68,13 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
-
+      <dependency>
```
ahamdi, line 71: Could you import the parent POM of ECMS instead ?

documents-storage-jcr/src/test/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorageTest.java
```diff
@@ -185,8 +189,11 @@ public void duplicateDocument() throws Exception {
     when(currentNode.addNode("copy of test","nt:file")).thenReturn(currentNode);
     when(identity.getRemoteId()).thenReturn("username");
     when(JCRDocumentsUtil.getUserSessionProvider(repositoryService, userID)).thenReturn(sessionProvider);
+    when(CommonsUtils.getService(AutoVersionService.class)).thenReturn(autoVersionService);
+    when(autoVersionService.autoVersion(any(Node.class))).thenReturn(null);
```
ahamdi: This instruction not needed IMO, you will just spy on the autoVersionService to check if the function was called, it won't be called really.

ahamdi (changes requested):

pom.xml
GouadriaHanen: done!

documents-storage-jcr/src/test/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorageTest.java
GouadriaHanen: done

ahamdi (approved):

pom.xml
```diff
@@ -67,7 +68,13 @@
...
+        <groupId>org.exoplatform.ecms</groupId>
+        <artifactId>ecms</artifactId>
+        <version>${org.exoplatform.ecms.version}</version>
```
hbenali, line 74: Wrong property name Use `addon.exo.ecms.version` instead. cc @ahamdi

hakermi:
Ecms dependancy is not allowed in documents app

PR #632, opened 2022-12-16
fix: wrong space document after creating a document shortcut - EXO-61289

GouadriaHanen:
prior to this change, after creating a shortcut for another space the shortcut is successfully created under the folder, but the current space is displayed instead of the space where I have created the shortcut.
After this change, redirection for the chosen space after creating a shortcut and the specific shortcut is opened.

### DIFF
--- a/documents-webapp/src/main/webapp/vue-app/documents/components/DocumentsMain.vue
+++ b/documents-webapp/src/main/webapp/vue-app/documents/components/DocumentsMain.vue
@@ -430,10 +430,6 @@ export default {
         }
       }
     },
-    openCreatedSpaceShortcut(parentFolder, space){
-      const folderPath = parentFolder?.path.split('/Documents/')[1];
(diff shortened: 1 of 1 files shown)


documents-webapp/src/main/webapp/vue-app/documents/components/DocumentsMain.vue
```diff
@@ -430,6 +430,10 @@ export default {
         }
       }
     },
+    openCreatedSpaceShortcut(parentFolder, space){
```
ahamdi: No need to add a new function if it will be used once, could you put it inside the if/else block after the shortcut creation ?

ahamdi (changes requested):

documents-webapp/src/main/webapp/vue-app/documents/components/DocumentsMain.vue
GouadriaHanen: done!

ahamdi (approved):

PR #638, opened 2022-12-20
fix: Elarge document content width - EXO-54759 .

akhanfir:
Prior to this change, document content width as not notes. After this change, document and notes have same content width.


documents-webapp/src/main/webapp/vue-app/documents/main.js
```diff
@@ -42,6 +42,9 @@ export function init() {
   if ( !document.getElementById('UIJcrExplorerContainer').classList.contains('VuetifyApp') ){
     document.getElementById('UIJcrExplorerContainer').classList.add('VuetifyApp');
   }
+  if (document.getElementById('UIJCRExplorerPortlet')) {
+    document.getElementById('UIJCRExplorerPortlet').style = 'padding: 20px 0px !important';
```
ahamdi, line 46: Is there a reason to add styles here ? couldn't we use CSS files for this change ?

hbenali:
Stale PR

PR #655, opened 2023-01-02
fix: Documents dissappear when filtering them by name, date or visibility - EXO-61422

akhanfir:
Before this change, when shared a document from another space and you sort files by name, some files disappear or if you sort them by date or visibility, all files disappear. After this change, all documents uploaded/created/symlinks are displayed in the correct selected order.


ahamdi (changes requested):
Could you a Unit test to cover this fix ? thanks

ahamdi (approved):

PR #676, opened 2023-01-12
fix: Clicking on open in documents should open the file's path not the docement preview - EXO-61531

GouadriaHanen:
before this change, in document preview when clicking on open in documents, it is redirected to the file folder and reopened in the docs view, instead of opening the file's path.
after this change, the file path is opened

### DIFF
--- a/documents-webapp/src/main/webapp/vue-app/documents/components/DocumentsMain.vue
+++ b/documents-webapp/src/main/webapp/vue-app/documents/components/DocumentsMain.vue
@@ -207,7 +207,7 @@ export default {
     this.$root.$on('documents-open-attachments-drawer', this.openDrawer);
     this.$root.$on('documents-filter', filter => {
       this.primaryFilter = filter;
-      this.refreshFiles(this.primaryFilter);
+      this.refreshFiles({'primaryFilter': this.primaryFilter});
     });
     this.$root.$on('show-alert', message => {
       this.displayMessage(message);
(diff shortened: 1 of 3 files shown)


documents-webapp/src/main/webapp/vue-app/documents/components/DocumentsMain.vue
```diff
@@ -216,7 +216,9 @@ export default {
       .finally(() => {
         this.checkDefaultViewOptions();
         this.optionsLoaded = true;
-        this.refreshFiles()
+        const queryParams = new URLSearchParams(window.location.search);
+        const disablePreview = queryParams.has('path');
+        this.refreshFiles(null, null, null, null, null, disablePreview)
```
ahamdi: Could you refactor the function refreshFiles to take an object as parameter intead of the current 6 parameters ?

ahamdi (changes requested):

documents-webapp/src/main/webapp/vue-app/documents/components/body/views/DocumentsFolderView.vue
```diff
@@ -172,7 +172,7 @@ export default {
     this.$root.$on('documents-filter', this.updateFilter);
   },
   mounted(){
-    this.$documentsUtils.injectSortTooltip(this.$t('documents.sort.tooltip'),'tooltip-marker');
+    this.$documentsUtils?.injectSortTooltip(this.$t('documents.sort.tooltip'),'tooltip-marker');
```
ahamdi: Could you remove this as it was fixed by https://github.com/exoplatform/documents/commit/a65c3f0d68f884ddb1a1c4303132abde01939a4a

GouadriaHanen: done!

ahamdi (approved):

PR #720, opened 2023-01-30
fix: deleted files reappear - EXO-61938

akhanfir:
Prior to this change, when open doc app of SpaceX and upload 2 files after delete them, deleted files reappear.
After this modification, the deleted file will not reappear after deletion.


documents-webapp/src/main/webapp/vue-app/documents/components/DocumentsMain.vue
```diff
@@ -586,6 +590,15 @@ export default {
...
         .finally(() => this.loading = false);
     },
+    isDocumentsToBeDeleted(doc) {
+      let result = true;
+      this.documentsToBeDeleted.forEach(documentId => {
+        if (doc.id === documentId) {
+          return result = false;
+        }
+      });
+      return result;
```
ahamdi:
```suggestion
      this.documentsToBeDeleted.forEach(documentId => {
        if (doc.id === documentId) {
          return false;
        }
      });
      return true;
```

ahamdi (changes requested):
What happens if we refresh the page ?

ahamdi:
@akhanfir what happens if the page is refreshed ?

ahamdi (approved):

PR #747, opened 2023-02-08
fix: fix the 3 dots menu actions are not immediatly displayed after the creation of a folder - EXO-62146

IlhemEssaadi:
Prior this change, not all actions in the 3 dots menu are available after creating a folder. After this change, all actions are available, this has been fixed by adding the value of the canAdd attribute to the newly created folder.

### DIFF
--- a/documents-webapp/src/main/webapp/vue-app/documents/components/DocumentsMain.vue
+++ b/documents-webapp/src/main/webapp/vue-app/documents/components/DocumentsMain.vue
@@ -702,6 +702,7 @@ export default {
       const ownerId = eXo.env.portal.spaceIdentityId || eXo.env.portal.userIdentityId;
       this.$documentFileService.createFolder(ownerId,this.parentFolderId,this.folderPath,name)
         .then(createdFolder => {
+          createdFolder.canAdd = this.canAdd;
           this.files.shift();
           this.files.unshift(createdFolder);
         }).catch(e => {


ahamdi (dismissed):

ahamdi (approved):

ahamdi (changes requested):
After a second thought looking at the source code, could you avoid reloading files from backend ?

documents-webapp/src/main/webapp/vue-app/documents/components/DocumentsMain.vue
```diff
@@ -702,8 +702,10 @@ export default {
       const ownerId = eXo.env.portal.spaceIdentityId || eXo.env.portal.userIdentityId;
       this.$documentFileService.createFolder(ownerId,this.parentFolderId,this.folderPath,name)
         .then(createdFolder => {
+          createdFolder.canAdd = this.canAdd;
           this.files.shift();
           this.files.unshift(createdFolder);
+          this.refreshFiles();
```
ahamdi: Couldn't we avoid this call ?

ahamdi (changes requested):

documents-webapp/src/main/webapp/vue-app/documents/components/DocumentsMain.vue
IlhemEssaadi: > Couldn't we avoid this call ?

Done !

ahamdi (approved):

PR #748, opened 2023-02-08
fix: fix restore the deleted code related to move document - EXO-62279

IlhemEssaadi:
After this change, we restore the deleted code related to the method of moving the document.


ahamdi (changes requested):
Could you update the title and description of the pull request to show what will be fixed by the PR ?

ahamdi (approved):

PR #752, opened 2023-02-13
fix: In document app, when clicking on download in action menu, with a large file, the click on download button do nothing - EXO-62303

rdenarie:
Before this fix, when the user clicks on download button in document application, the item was fetch by javascript. In case of large item, it could be very long to execute This fix modify the behaviour, so that the user directly download the item. there is no more javascript fetch

### DIFF
--- a/documents-webapp/src/main/webapp/vue-app/documents/components/body/actions/DownloadMenuAction.vue
+++ b/documents-webapp/src/main/webapp/vue-app/documents/components/body/actions/DownloadMenuAction.vue
@@ -40,31 +40,19 @@ export default {
           if (urlDownload.indexOf('/') > 0 && !urlDownload.includes(window.location.hostname)) {
             return;
           }
-          return fetch(urlDownload, {
-            credentials: 'include',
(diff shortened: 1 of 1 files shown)


documents-webapp/src/main/webapp/vue-app/documents/components/body/actions/DownloadMenuAction.vue
```diff
@@ -29,8 +29,10 @@ export default {
   },
   methods: {
     download() {
+      console.log(this.file.id);
```
ahamdi: could you remove the console.log statements ?

ahamdi (changes requested):

documents-webapp/src/main/webapp/vue-app/documents/components/body/actions/DownloadMenuAction.vue
rdenarie: done

ahamdi (approved):

PR #756, opened 2023-02-14
fix: bad display for filter badge icon in mobile vue - EXO-62361

GouadriaHanen:
prior to this change, The filter option button is too big and the number is displayed under the icon after this change, The number is displayed in the same line with the icon


ahamdi:
@GouadriaHanen @halaya @aycherif This job is using JDK 11, and it should use java 17 instead

ahamdi (approved):

PR #780, opened 2023-02-24
fix : Enable to move a folder in a certain cases - EXO-62663

sofyenne:
To sort folders, we use `NumberUtils.isNumber()` library to check if the folder or file name is a number. 
However in some cases we may have a specific document name like  `12f` or `12L` which is considered a valid number in java , as result the `isNumber` method return true but if we try to parse this number we received a `numberFormatException` because this number format cannot be parsed , therefor the tree-view of the move to drawer not displayed  if we want to move folder with this specific name.
To handle this exception we need to replace the `isNumber` method with `isParsable` method.

### DIFF
--- a/documents-storage-jcr/src/test/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorageTest.java
+++ b/documents-storage-jcr/src/test/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorageTest.java
@@ -1,5 +1,6 @@
 package org.exoplatform.documents.storage.jcr;
 
+import org.apache.commons.lang3.math.NumberUtils;
 import org.exoplatform.commons.ObjectAlreadyExistsException;
 import org.exoplatform.commons.utils.CommonsUtils;
 import org.exoplatform.documents.constant.DocumentSortField;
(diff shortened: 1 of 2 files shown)


ahamdi (changes requested):
Coud you add a Unit test for this change ?

sofyenne:
> Coud you add a Unit test for this change ?

Done !

documents-storage-jcr/src/test/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorageTest.java
```diff
@@ -302,6 +319,34 @@ public void getFolderChildNodes() throws Exception {
...
+                            "toNodes",
+                            identityManager,
+                            userSession,
+                            nodeIterator2,
+                            identity,
+                            spaceService,
+                            false);
+    assertEquals(NumberUtils.isParsable(folderWithNumericName.getName()),true);
+    assertEquals(NumberUtils.isParsable(folderWithSpecificName.getName()),false);
+    assertEquals(NumberUtils.isParsable(folderWithSpecificName1.getName()),false);
```
ahamdi: No need to test ```NumberUtils.isParsable()```

sofyenne: done !

ahamdi (approved):

PR #781, opened 2023-02-28
fix : favorite quick filter doesn't work on document application - EXO-62473

sofyenne:
On the timeline view of the personal drive, when we create a new file and add it to our favorites, and then select the "favorites" quick filter, the newly added file is not displayed. The problem is that the "favorites" quick filter only returns files inside the Private route path, while we need it to return files inside both the Public and Private paths.
This change addresses the issue.


documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorage.java
```diff
@@ -167,6 +167,9 @@ public List<FileNode> getFilesTimeline(DocumentTimelineFilter filter,
         String workspace = session.getWorkspace().getName();
         String sortField = getSortField(filter, false);
         String sortDirection = getSortDirection(filter);
+        if (rootPath.endsWith(username + "/Private")) {
```
ahamdi, line 170: I am not sure we need this change, the Private folder has already a symlink to the /Public folder. it should be another cause for filtering files from /Public files, could you check again ?

ahamdi (changes requested):

PR #782, opened 2023-02-28
fix : Enable to move a folder in a certain cases - EXO-62663 

sofyenne:
To sort folders, we use `NumberUtils.isNumber()` library to check if the folder or file name is a number. However in some cases we may have a specific document name like `12f` or `12L` which is considered a valid number in java , as result the `isNumber` method return true but if we try to parse this number we received a `numberFormatException` because this number format cannot be parsed , therefor the tree-view of the move to drawer not displayed if we want to move folder with this specific name. To handle this exception we need to replace the `isNumber` method with `isParsable` method.


ahamdi (commented):
Could you fix the code smells ?

sofyenne:
> Could you fix the code smells ?

fixed !

ahamdi (approved):

PR #800, opened 2023-03-08
fix : Disable the ability to open the access manage drawer when the sharing of documents is suspended  - EXO-62870

sofyenne:
This change is going to disable the management access icon from begin clickable when the sharing documents is suspended.

### DIFF
--- a/documents-webapp/src/main/webapp/vue-app/documents/components/body/table/cells/DocumentsVisibilityCell.vue
+++ b/documents-webapp/src/main/webapp/vue-app/documents/components/body/table/cells/DocumentsVisibilityCell.vue
@@ -16,7 +16,7 @@
         </v-icon>
       </v-btn>
     </template>
-    <span class="center">{{ icon.title }}</span>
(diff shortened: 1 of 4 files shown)


documents-webapp/src/main/webapp/vue-app/documents/components/body/table/cells/DocumentsVisibilityCell.vue
```diff
@@ -92,11 +93,19 @@ export default {
...
   },
+  created() {
+    this.$transferRulesService.getDocumentsTransferRules().then(rules => {
+      this.sharedDocumentSuspended = rules.sharedDocumentStatus === 'true';
+    });
+  },
   methods: {
     changeVisibility() {
-      if (!this.file.acl.canEdit) {
+      if (!this.file.acl.canEdit || this.sharedDocumentSuspended) {
```
ahamdi, line 108: Could you also check if the suspend is considered from backend ?

sofyenne: done !

ahamdi (approved):

PR #801, opened 2023-03-10
fix: Redirection issue when sharing a folder - EXO-63010

GouadriaHanen:
prior to this change, after sharing a folder when clicking on sharing notification  the user is not directed inside the shared folder since the URL redirection is only for attachments (documentPreviewId: for previewing attachments"
after this change, the URL is well built with documentPreviewId for attachments and folderId for folders

### DIFF
--- a/documents-services/src/main/java/org/exoplatform/documents/notification/utils/NotificationUtils.java
+++ b/documents-services/src/main/java/org/exoplatform/documents/notification/utils/NotificationUtils.java
@@ -18,11 +18,7 @@ package org.exoplatform.documents.notification.utils;
 
 import org.exoplatform.commons.utils.CommonsUtils;
 import org.exoplatform.documents.rest.util.EntityBuilder;
-import org.exoplatform.services.jcr.RepositoryService;
 import org.exoplatform.services.jcr.core.ExtendedNode;
-import org.exoplatform.services.jcr.core.ExtendedSession;
-import org.exoplatform.services.jcr.ext.app.SessionProviderService;
-import org.exoplatform.services.jcr.ext.common.SessionProvider;
 import org.exoplatform.services.jcr.impl.core.NodeImpl;
 import org.exoplatform.social.core.identity.model.Identity;
 import org.exoplatform.social.core.identity.model.Profile;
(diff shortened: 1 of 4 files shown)


documents-services/src/main/java/org/exoplatform/documents/notification/utils/NotificationUtils.java
```diff
@@ -52,7 +62,7 @@ public static String getDocumentLink(Node node, SpaceService spaceService, Ident
     return stringBuilder.toString() ;
   }
 
-  public static String getSharedDocumentLink(String nodeUuid, SpaceService spaceService, String spacePrettyName) {
+  public static String getSharedDocumentLink(Node sharedNode, SpaceService spaceService, String spacePrettyName) throws RepositoryException {
```
ahamdi, line 61: Could you check that this function is not used by other modules ?

documents-services/src/main/java/org/exoplatform/documents/notification/utils/NotificationUtils.java
```diff
@@ -96,4 +107,21 @@ public static Profile getUserProfile(IdentityManager identityManager, String use
...
+  public static boolean isNodeFile(Node node) {
+    try {
+      SessionProviderService sessionProviderService = CommonsUtils.getService(SessionProviderService.class);
+      RepositoryService repositoryService = CommonsUtils.getService(RepositoryService.class);
+      SessionProvider sessionProvider = sessionProviderService.getSessionProvider(null);
+      Session session = sessionProvider.getSession(
+                                                   repositoryService.getCurrentRepository()
+                                                                    .getConfiguration()
+                                                                    .getDefaultWorkspaceName(),
+                                                   repositoryService.getCurrentRepository());
```
ahamdi: could you use node.deSession() ?

GouadriaHanen: done!

documents-services/src/main/java/org/exoplatform/documents/notification/utils/NotificationUtils.java
GouadriaHanen: checked and used only in documents module

ahamdi (approved):

PR #802, opened 2023-03-13
fix: fix documents lost after shortcut and move actions - EXO-63061

IlhemEssaadi:
Prior to this fix, when we move and create a new version of a document that has a shortcut in the same place, we lose the documents and the newly created documents are not displayed, which is due to the deletion of the source node of the shortcut.
After this change, the version is created and displayed in the dedicated folder and all newly created documents are displayed.


ahamdi (commented):
Could you add a unit test ?

ahamdi (dismissed):
LGTM

ahamdi (approved):

PR #804, opened 2023-03-13
fix: [Document] sort rules are not respected during filter by name and extended search - EXO-62937.

akhanfir:
Prior to this change,the filter by name in the list of document files and extended search is not working properly. After this changes, this list of document files displays the workflows correctly.

### DIFF
--- a/documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/search/DocumentSearchServiceConnector.java
+++ b/documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/search/DocumentSearchServiceConnector.java
@@ -117,7 +117,18 @@ public class DocumentSearchServiceConnector {
     if (StringUtils.isBlank(filter.getQuery()) && !filter.getFavorites()) {
       throw new IllegalArgumentException("Filter term is mandatory");
     }
-    String esQuery = buildQueryStatement(userIdentity, workspace, path, filter, offset, limit);
+    switch (sortField) {
+      case "lastUpdatedDate" :
+        sortField = "lastUpdatedDate";
+        break;
+      case "title" :
+        sortField = "title.raw";
+        break;
+      default:
+        sortField = "_score";
+    }
+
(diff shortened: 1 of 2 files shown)


rdenarie (changes requested):
The pr have conflict, can you update ?

Can you also rework the commit message ?
In your message, you have this 

> Prior to this change,the filter by name in the list of document files and extended search is not working properly

It does not explain WHY the filter not working. Please try to answer question WHY

and this :

> After this changes, this list of document files displays the workflows correctly.

It does not explain HOW you fix the problem. Please try to reword by answering the question HOW.

Finally, you try to fix the problem by sorting the search result list, after the search. 
But, in documentSearchServiceConnector, we pass options sortField and sortDirection. So the result list should be sorted by the search connector. 

In addition, my IDE show this 
image

The sortField and sortDirection are not used in the search function. I think it could be better to use it in the ES query, so that ES is in charge of sorting the result, and we get already sorted results, no need to recompute it.

exo-swf:

rdenarie (changes requested):
Can you rebase the pr

exo-swf:

exo-swf:

documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/search/DocumentSearchServiceConnector.java
```diff
@@ -117,7 +117,11 @@ public Collection<SearchResult> search(Identity userIdentity,
     if (StringUtils.isBlank(filter.getQuery()) && !filter.getFavorites()) {
       throw new IllegalArgumentException("Filter term is mandatory");
     }
-    String esQuery = buildQueryStatement(userIdentity, workspace, path, filter, offset, limit);
+    if (!sortField.equalsIgnoreCase("lastUpdatedDate")) {
+      sortField="_score";
+    }
```
rdenarie, line 129: if sortField is title, you must sort on title.raw

rdenarie (changes requested):

rdenarie (approved):

PR #805, opened 2023-03-14
 fix:[Document] Loading area doesn't disappear when canceling drag and drop operation - EXO-62932

akhanfir:
Prior to this change, when try to drag and drop a document and cancel the operation before adding the document, The loading area doesn't disappear and can't be closed. After this change, The loading area disappear when we are not hovering over the document/folder list with a document.

### DIFF
--- a/documents-webapp/src/main/webapp/vue-app/documents/components/DocumentsMain.vue
+++ b/documents-webapp/src/main/webapp/vue-app/documents/components/DocumentsMain.vue
@@ -4,131 +4,133 @@
     :class="isMobile ? 'mobile' : ''"
     role="main"
     flat>
-    <div
-      class="pa-3 white"
-      @dragover.prevent
-      @drop.prevent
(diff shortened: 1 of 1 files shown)


documents-webapp/src/main/webapp/vue-app/documents/components/DocumentsMain.vue
```diff
@@ -1076,17 +1079,25 @@ export default {
...
       //folder haven't type
       const isFile = item && item.type !== '';
       if (this.canAdd && isFile){
         this.$root.$emit('show-upload-overlay');
       }
     },
+    hideOverlay() {
+      if (this.showOverlay){
+        this.showOverlay = false;
+        this.$root.$emit('hide-upload-overlay',this.hide);
```
ahamdi: what is the property this.hide ?

ahamdi (approved):

exo-swf:

PR #814, opened 2023-03-21
fix : Adding documents and folders in a read only shared folder should not be possible - EXO-63084

sofyenne:
This change is going to disallow the ability to add a new folder inside a shared folder with read only permission

### DIFF
--- a/documents-webapp/src/main/webapp/vue-app/documents/components/header/actions/DocumentsAddNewFile.vue
+++ b/documents-webapp/src/main/webapp/vue-app/documents/components/header/actions/DocumentsAddNewFile.vue
@@ -97,7 +97,7 @@ export default {
       return this.selectedView === 'folder';
     },
     disableButton(){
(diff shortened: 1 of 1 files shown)


rdenarie (changes requested):
In the task, the expected behaviour mention this 

> The user should not be authorized on backend to upload/create document and add folder into a readonly shared folder

As I can see, only the front part was fixed.
Do we have something to do on backend ?

sofyenne:
> In the task, the expected behaviour mention this
> 
> > The user should not be authorized on backend to upload/create document and add folder into a readonly shared folder
> 
> As I can see, only the front part was fixed. Do we have something to do on backend ?


> In the task, the expected behaviour mention this
> 
> > The user should not be authorized on backend to upload/create document and add folder into a readonly shared folder
> 
> As I can see, only the front part was fixed. Do we have something to do on backend ?

 @rdenarie  This PR is a correction after the test. The fix for the backend part has been merged.

rdenarie:
Ok, thanks for the precision.
So it is ok for me

rdenarie (approved):

PR #816, opened 2023-03-21
fix: Document not visible in recent view after empty the description - EXO-63280

hakermi:
Prior to this change, After creation a document a fill the description and empty it again, the document disappears from recent view because of a missing check on description property leads to and `ArrayOutOfBoundException`.
This PR ensures that the multi-valued description property array is not empty before accessing its elements


rdenarie (changes requested):
I'm not sure that the unit test is relevant.

As you modify in JCRDocumentUtils, you should add/update a test in JCRDocumentUtilsTest, in which you already have a test named testRetrieveFileProperties

You should create a test in which the node have the jcr property, which is empty. Then, without the fix, the test fails. With the fix, the test succeed

rdenarie (changes requested):
Ok for the new test.
Should we remove the updateDocumentDescription test ?

hakermi:
> Ok for the new test. Should we remove the updateDocumentDescription test ?

It was something that had to be tested, and now we had the opportunity to add the unit test and increase test coverage.

rdenarie (approved):
so ok for me, thanks for the clarification

PR #832, opened 2023-03-23
fix: load correctly contents of folder : subfolders then files with symlinks included - EXO-62645

ahamdi:
When displaying contents of a folder in the Documents application, this fix will reorder folders & files retrieved from backend by sending first folders and symlinks of folders, then files and symlinks of files.


GouadriaHanen (dismissed):

rdenarie:
There is a build problem

sofyenne (commented):
@ahamdi we will merge this PR after the PR of transfer rules :
https://github.com/exoplatform/documents/pull/834

GouadriaHanen (approved):

PR #841, opened 2023-03-27
fix: Edit option is no more present in 3 dots menu

rdenarie:
Before this fix, extensions about supported mimetype is loaded after extensions about 3 dots menu, so, for edit action, the supportedMimeType is empty, and so, the file is not editable

This fix ensure to load mimetype extension before 3 dots menu


ahamdi:
@rdenarie Build is failing, could you check ?

rdenarie:
Yes, I have the modification locally, and I need to push, but there is a githuib outage.
Ill push as soon as possible

rdenarie:
Yes, I have the modification locally, and I need to push, but there is a githuib outage.
Ill push as soon as possible

PR #848, opened 2023-03-30
feat: Multi selection menu action for desktop and mobile (UI) - EXO-59572,EXO-59698

hakermi:
Multi selection menu action ui for desktop and mobile


mkrout (approved):
LGTM

PR #854, opened 2023-04-04
 fix: [Document] Can't disable document edition on which there are all rights - EXO-63201

akhanfir:
Prior to this change, when open the manage access drawer of file then switch the button 'Allow everyone to edit and saves', an error message 'Update of the access failed' displays and the manage access drawer is forever loading. To fix that, save the node after set the `dateModified` and `lastModifiedDate` in the function `updatePermissions()`. after this change, the document is modifiable and no error message is displayed .

### DIFF
--- a/documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorage.java
+++ b/documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorage.java
@@ -1106,11 +1106,11 @@ public class JCRDocumentFileStorage implements DocumentFileStorage {
       if (node.canAddMixin(NodeTypeConstants.EXO_PRIVILEGEABLE)) {
         node.addMixin(NodeTypeConstants.EXO_PRIVILEGEABLE);
       }
+      Calendar now = Calendar.getInstance();
+      node.setProperty(NodeTypeConstants.EXO_DATE_MODIFIED, now);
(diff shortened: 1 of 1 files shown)


documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorage.java
```diff
@@ -1106,11 +1106,11 @@ public void updatePermissions(String documentID, NodePermission nodePermissionEn
       if (node.canAddMixin(NodeTypeConstants.EXO_PRIVILEGEABLE)) {
         node.addMixin(NodeTypeConstants.EXO_PRIVILEGEABLE);
       }
+      Calendar now = Calendar.getInstance();
+      node.setProperty(NodeTypeConstants.EXO_DATE_MODIFIED, now);
+      node.setProperty(NodeTypeConstants.EXO_LAST_MODIFIED_DATE, now);
+      node.save();
```
ahamdi, line 1112: The issue was that the changes contain opposite instructions, that means the user removes possibility to update file and tries to update the lastModification date in the same time which will execute the permission removal , then it will try to update the modification date.
Removing this block should fix the issue. could you test that ?

ahamdi: @akhanfir No need to remove the block, let's keep it

ahamdi (approved):

PR #866, opened 2023-04-12
fix: Fix visibility rules when share folder with a user personal drive - EXO-63701

sofyenne:
Before this change, when we shared a folder  with a user personal drive and gave read-only permissions, the folder would be created but the visibility rules were incorrect. The problem was that when we added a symlink to the destination user personal drive , the symlink was added with all permissions, allowing it to open the visibility drawer.

With this fix, we will add the symlink to the destination space with the provided permissions and correct the wrong visibility rule icon.

### DIFF
--- a/documents-storage-jcr/src/test/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorageTest.java
+++ b/documents-storage-jcr/src/test/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorageTest.java
@@ -176,7 +176,10 @@ public class JCRDocumentFileStorageTest {
     when(currentNode.getPrimaryNodeType()).thenReturn(nodeType);
     when(nodeType.getName()).thenReturn("nt:file");
     when(((ExtendedNode) currentNode).getIdentifier()).thenReturn("123");
+    AccessControlEntry accessControlEntry = new AccessControlEntry("username", PermissionType.READ);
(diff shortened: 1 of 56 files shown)


documents-storage-jcr/src/test/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorageTest.java
```diff
@@ -176,7 +176,10 @@ public void shareDocument() throws Exception {
     when(currentNode.getPrimaryNodeType()).thenReturn(nodeType);
     when(nodeType.getName()).thenReturn("nt:file");
     when(((ExtendedNode) currentNode).getIdentifier()).thenReturn("123");
+    AccessControlEntry accessControlEntry = new AccessControlEntry("username", PermissionType.READ);
     when(identity.getRemoteId()).thenReturn("username");
+    AccessControlList acl1 = new AccessControlList("username", Arrays.asList(accessControlEntry));
+    when(((ExtendedNode) currentNode).getACL()).thenReturn(acl1);
```
ahamdi, line 180: Could you add more checks to make sure the permission are well considered ? i.e when we share with read-only and when we share with read-write

sofyenne: Done !

ahamdi (approved):

PR #871, opened 2023-04-13
fix: Sort tree-view items on move to drawer - EXO-63141

sofyenne:
Before to this change when opening move to or create shortcut drawer the tree-view items displayed not sorted  ( folder 10 displayed before folder 2  ). The problem was that the comparison between the two names is based on subtraction, which may not always produce the correct ordering . For example, when comparing "2" and "10", the result of the subtraction would be -8, which indicates that "2" comes before "10". However, when comparing "10" and "2", the result of the subtraction would be 8, which indicates that "10" comes before "2"

To fix this, we use the Integer.compare() method to compare the numeric values of the names instead of subtraction.


ahamdi:
@sofyenne Integer Comparison is based on subtraction too , isn't it ?

sofyenne:
> @sofyenne Integer Comparison is based on subtraction too , isn't it ?

@ahamdi yes it is  , but using Integer.compare(number1, number2) will produce the correct ordering, whereas using subtraction may not.

ahamdi (approved):

PR #888, opened 2023-04-24
fix: share document with group permissions - EXO-63772

sofyenne:
Prior to this change, when we wanted to share a document with a group permission, the document would not be shared. This change is going to implement the ability to share documents with group permission where it was missing.


exo-swf:

PR #894, opened 2023-04-26
fix : Blank content page displayed when open document app - EXO-63718

sofyenne:
Before to this change when we opened the document application a blank page content would be displayed with a console error 'Cannot read properties of undefined (reading 'getDocumentsTransferRules')'. 
This change is going to pass a callback function to Vue.prototype.$transferRulesService, which will be executed after the next DOM update cycle.


ahamdi (approved):
This is a temporary solution until the attachment app & service is moved to a common package

PR #904, opened 2023-05-03
fix : After collaborating on a document group permission isn't displayed in the access manage drawer  -EXO-63772

sofyenne:
Before to this change when collaborating on a document with a group permission by adding the group as collaborators from the manage access drawer the document permission wouldn't be updated and the group  wouldn't be displayed on the manage access drawer , This change will update the  permission update method to allow the addition of  the provided group permission to the document's permission and create an identity model for the group , similar to the one built by the identity suggester in order to display it in the collaborators suggester list .


documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/util/JCRDocumentsUtil.java
```diff
@@ -477,6 +481,8 @@ public static void computeDocumentAcl(Node node, AbstractNode documentNode, Iden
           if(identity!=null){
             permissions.add(new PermissionEntry(identity, accessControlEntry.getPermission(),getPermissionRole(accessControlEntry.getMembershipEntry().getMembershipType())));
           }
+        } else if (!membershipEntry.toString().equals(GROUP_ADMINISTRATORS) && groupToIdentity(membershipEntry.getGroup()) != null) {
```
ahamdi: Why are you checking on the Administrators group ?

documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorage.java
```diff
@@ -1112,6 +1112,14 @@ public void updatePermissions(String documentID, NodePermission nodePermissionEn
               permissions.put("redactor:" + groupId, new String[] { PermissionType.READ });
             }
           }
+        } else if (permission.getIdentity().getProviderId().equals("group")) {
```
ahamdi, line 1115: we need UT for this change

sofyenne: done !

exo-swf:

documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/util/JCRDocumentsUtil.java
ahamdi: @sofyenne could you answer my question here !

sofyenne: > Why are you checking on the Administrators group ?

By default, the Administrators group is included as collaborators on the document, so there is no need to display it on the document's collaborators list

ahamdi: @sofyenne What if it is not the case ? I mean permissions could be altered and this group is not always added as collaborators in folders.
I know that by default administrators have access to space documents, but they do not have access to personal ones for example , and we can remove them whenever we want from space documents.

sofyenne: @ahamdi  No problem if the  administrators group is not a collaborator; we check if they are a collaborator in order to not display them on the list .

ahamdi: @sofyenne we are not checking wether they are collaborator here, I prefer to remove this check
I don't think there is a requirement to not displaying the administrators grou from the list of collaborators, isn't it ?

sofyenne: Done!

ahamdi (approved):

PR #920, opened 2023-05-18
fix : Enable to upload document in folder with name contains '.' character  - EXO-64403 - EXO-64406

sofyenne:
Before to this change we weren't able to create a folder with name containing a '.' followed by a special character  , the issue  stemmed to the cleanName method which uses the '.' character to identify the file's extension and remove it from the name , so if we have a folder name like  `test.test#test` the last part after the '.' character wouldn't be cleaned ,  resulting this problem .

To address this issue, we will update the method cleanName to check the document type before cleaning the name of document.

### DIFF
--- a/documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorage.java
+++ b/documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorage.java
@@ -595,7 +595,7 @@ public class JCRDocumentFileStorage implements DocumentFileStorage {
       }
       //no need to this object later make it eligible to the garbage collactor
       nodeAccessList = null;
-      String name = Text.escapeIllegalJcrChars(cleanName(title.toLowerCase()));
+      String name = Text.escapeIllegalJcrChars(cleanFolderName(title.toLowerCase()));
       if (node.hasNode(name)) {
         throw new ObjectAlreadyExistsException("Folder'" + name + "' already exist");
       }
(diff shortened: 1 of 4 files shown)


documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorage.java
```diff
@@ -595,7 +595,7 @@ public AbstractNode createFolder(long ownerId,
       }
       //no need to this object later make it eligible to the garbage collactor
       nodeAccessList = null;
-      String name = Text.escapeIllegalJcrChars(cleanName(title.toLowerCase()));
+      String name = Text.escapeIllegalJcrChars(cleanFolderName(title.toLowerCase()));
```
ahamdi: Why should we create a new function for folders, let's use the same one and clean both parts for files and folders .

ahamdi (changes requested):

exo-swf:

documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorage.java
sofyenne: done !

sofyenne: done!

documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/util/JCRDocumentsUtil.java
```diff
@@ -633,10 +634,10 @@ public static long getUserIdentityId(IdentityManager identityManager, String use
    *
    * @return the string
    */
-  public static String cleanName(String oldName) {
+  public static String cleanName(String oldName, String nodeType) {
```
ahamdi, line 639: I think that the API is used by other modules , we can not change it without checking its usages.
I prefer to re-create the old function  **public static String cleanName(String oldName)** that will call this one with **NodeTypeConstants.NT_FILE** as the nodetype by default

ahamdi (changes requested):

ahamdi (approved):

PR #926, opened 2023-05-22
fix: Prevent group administrators from being displayed in the access management drawer for non-group administrator members - EXO-64412

sofyenne:
This change is going to prevent group administrators from being displayed in the access management drawer for non-group administrator members .

### DIFF
--- a/documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/util/JCRDocumentsUtil.java
+++ b/documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/util/JCRDocumentsUtil.java
@@ -480,6 +480,9 @@ public class JCRDocumentsUtil {
             permissions.add(new PermissionEntry(identity, accessControlEntry.getPermission(),getPermissionRole(accessControlEntry.getMembershipEntry().getMembershipType())));
           }
         } else if (groupToIdentity(membershipEntry.getGroup()) != null) {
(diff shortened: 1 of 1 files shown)


documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/util/JCRDocumentsUtil.java
```diff
@@ -480,6 +480,9 @@ public static void computeDocumentAcl(Node node, AbstractNode documentNode, Iden
             permissions.add(new PermissionEntry(identity, accessControlEntry.getPermission(),getPermissionRole(accessControlEntry.getMembershipEntry().getMembershipType())));
           }
         } else if (groupToIdentity(membershipEntry.getGroup()) != null) {
+          if (membershipEntry.toString().equals("*:/platform/administrators") && !aclIdentity.isMemberOf(membershipEntry)) {
```
ahamdi:
```suggestion
          UserACL userACL = CommonsUtils.getService(UserACL.class);
          if (membershipEntry.getGroup().equals(userACL.getAdminGroups()) && !aclIdentity.isMemberOf(membershipEntry)) {
```

ahamdi (changes requested):

documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/util/JCRDocumentsUtil.java
sofyenne: done !

ahamdi (approved):

PR #942, opened 2023-05-25
fix: root folder in treeview drawer not translated EXO-64528

hthabti:
prior to this change, the root folder of treeview drawer wasn't translated. 
after this change it would be well translated


documents-webapp/src/main/webapp/vue-app/documents/components/body/views/FolderTreeViewDrawer.vue
```diff
@@ -25,7 +25,7 @@
             <v-list-item-title 
               class="body-2 mx-2 mt-1"
               :class="idItemActive === item.id ? 'primary--text font-weight-bold' : ''">
-              {{ item.name }}
+              {{ item.name === 'Documents' ? $t('documents.label.spaceHomeDocuments') : item.name }}
```
ahamdi: let's avoid expression inside the template if possible, could you add a computed property ?

ahamdi (changes requested):

documents-webapp/src/main/webapp/vue-app/documents/components/body/views/FolderTreeViewDrawer.vue
hthabti: I used method instead, it can't be done through computed (it is an item from array not from data, props..)

ahamdi (approved):

exo-swf:

PR #963, opened 2023-06-01
feat: Add document views detail - EXO-64500

hakermi:
Add document views detail which will display the number of views received by a document


exo-swf:

mkrout (approved):

PR #989, opened 2023-06-15
fix : Enable to search document with name contains an elastic search reserved characters - EXO-64651

sofyenne:
Before this change, when we were searching for a document with a name containing the "-" character, no results were displayed. The problem was that when we escaped the reserved characters in the search term, the "*" character in the search query was not escaped.

This change will verify if the search term contains any escaped reserved characters and then escape the "*" character in the search query.


ahamdi (commented):
Could you fix the SONAR issue and add a unit test for this change ? thanks

ahamdi:
could you add a UT too ? thanks

exo-swf:

ahamdi (approved):

PR #1005, opened 2023-06-26
fix : Fix the infinite loop that occurs when creating a shortcut of a folder within the same folder -EXO-65086

sofyenne:
Before this change, when we created a shortcut of a folder within the same folder, an infinite loop was thrown. This issue occurred due to the symlink check, which returned the symlink of the node itself. This change is going to resolve this problem by disallowing the addition of the same symlink to the breadcrumb list.

### DIFF
--- a/documents-api/pom.xml
+++ b/documents-api/pom.xml
@@ -3,7 +3,7 @@
   <parent>
     <groupId>org.exoplatform.documents</groupId>
     <artifactId>documents-parent</artifactId>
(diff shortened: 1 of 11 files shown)


ahamdi:
@sofyenne Is this a final fix or just a WA to solve the urgent issue ?

sofyenne:
> @sofyenne Is this a final fix or just a WA to solve the urgent issue ?

It is just to resolve the urgent issue (infinite loop)

ahamdi (approved):

PR #1013, opened 2023-07-03
feat: Remove the reminders of switch to old app and the introduction to the new app - EXO-65215

hakermi:
Remove the reminders of switch to old app and the introduction to the new app


mkrout (approved):
LGTM

PR #1033, opened 2023-07-11
fix: When a version is created by system, the version history drawer is empty - EXO-65435

rdenarie:
Before this fix, when a version is created by system session (by a job for example), the version history drawer is empty due to error in RestService : when computing user profile, the identity for system is not found, and so getting the profile create a null pointer exception This fix adress this special case to not have the NPE


ahamdi (approved):
LGTM

PR #1059, opened 2023-08-02
fix: Notify only corresponding collaborators after making changes on their permission on the document managed access drawer - EXO-65096

sofyenne:
Prior to this change, after adding a collaborator or updating the permissions of an existing collaborator in the document's managed access drawer, all collaborators  were notified that they were assigned as collaborators on the corresponding document. This issue was due to the updating of the permissions list of the linked node, even when we didn't have any changes, and broadcasting the shared document event to notify the destination identity.

With this change, before creating or updating the linked node's permission list, we will verify whether the linked node has already been created with the same permission list.

### DIFF
--- a/documents-storage-jcr/src/test/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorageTest.java
+++ b/documents-storage-jcr/src/test/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorageTest.java
@@ -191,14 +191,14 @@ public class JCRDocumentFileStorageTest {
     verify(linkNode).setPermissions(argThat((Map<String, String[]> map) -> map.containsKey("username") && Arrays.equals(map.get("username"),new String[]{"edit"})));
 
     when(rootNode.hasNode("Shared")).thenReturn(true);
-    when(rootNode.getNode("shared")).thenReturn(sharedNode);
+    when(rootNode.getNode("Shared")).thenReturn(sharedNode);
     when(sharedNode.hasNode(currentNode.getName())).thenReturn(true);
     when(sharedNode.getNode(currentNode.getName())).thenReturn(linkNode);
     when(linkNode.getACL()).thenReturn(acl);
 
     jcrDocumentFileStorage.shareDocument("1", 1L);
     // Assert that the shared document event was not broadcast
-    UTILS.verify(() -> Utils.broadcast(listenerService, "share_document_event", identity, linkNode), times(2));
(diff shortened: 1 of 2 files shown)


documents-storage-jcr/src/test/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorageTest.java
```diff
@@ -190,6 +190,17 @@ public void shareDocument() throws Exception {
...
     verify(linkNode).setPermissions(argThat((Map<String, String[]> map) -> map.containsKey("username") && Arrays.equals(map.get("username"),new String[]{"edit"})));
 
+    when(rootNode.hasNode("Shared")).thenReturn(true);
+    when(rootNode.getNode("shared")).thenReturn(sharedNode);
+    when(sharedNode.hasNode(currentNode.getName())).thenReturn(true);
+    when(sharedNode.getNode(currentNode.getName())).thenReturn(linkNode);
+    when(linkNode.getACL()).thenReturn(acl);
+
+    jcrDocumentFileStorage.shareDocument("1", 1L);
+    // Assert that the shared document event was not broadcast
```
ahamdi, line 200: I don't get it, the broadcast is done 2 times, but the comment says that it should not have been fired.

documents-storage-jcr/src/test/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorageTest.java
```diff
@@ -190,6 +190,17 @@ public void shareDocument() throws Exception {
     //assert that the linkNode set edit permission
     verify(linkNode).setPermissions(argThat((Map<String, String[]> map) -> map.containsKey("username") && Arrays.equals(map.get("username"),new String[]{"edit"})));
 
+    when(rootNode.hasNode("Shared")).thenReturn(true);
+    when(rootNode.getNode("shared")).thenReturn(sharedNode);
```
ahamdi:
```suggestion
    when(rootNode.getNode("Shared")).thenReturn(sharedNode);
```

documents-storage-jcr/src/test/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorageTest.java
sofyenne: The broadcast is performed twice in the previous test case. PR updated !

ahamdi (approved):

PR #1071, opened 2023-08-09
fix: Grant all permissions to space hosts and editors when the document visibility is set to Only designated collaborator -EXO-65224

sofyenne:
Before this change, when the document visibility was set to "Only designated collaborator," the document became inaccessible to the space hosts and redactors. This change is going to grant all permissions to space hosts and editors when the document visibility is set to Only designated collaborator .


ahamdi (approved):
LGTM

PR #1077, opened 2023-08-15
fix: Notify only corresponding collaborators after making changes on their permission on the document managed access drawer - EXO-65096

sofyenne:
After implementing this previous pull request https://github.com/exoplatform/documents/pull/1069 to notify only the corresponding collaborators after making changes to their permissions, space members are still receiving notifications.

With this new change, we will first verify whether the linked node has already been created with the same permission list before notifying the member.


ahamdi (commented):
Is there a UT dedicated for the both cases : when notifications are sent and when they are not

sofyenne:
> Is there a UT dedicated for the both cases : when notifications are sent and when they are not

UT updated !

ahamdi (approved):

PR #1084, opened 2023-08-22
fix : Remove the symlink when designating a document's collaborator - EXO-65859

sofyenne:
Prior to this change ,When a document was re-shared with a user (after being removed from the document collaborators list and reassigned), they did not receive a notification. This was because the symlink  already exists with the same permissions, but it was hidden from the collaborators by removing the permissions from the target node.
This change is going to remove the symlink from the shared folder if a user or a space is removed from the document collaborators list.

### DIFF
--- a/documents-services/src/main/java/org/exoplatform/documents/service/DocumentFileServiceImpl.java
+++ b/documents-services/src/main/java/org/exoplatform/documents/service/DocumentFileServiceImpl.java
@@ -331,7 +331,7 @@ public class DocumentFileServiceImpl implements DocumentFileService {
   }
 
   @Override
-  public void unShareDocument(String documentId, long destId) throws IllegalAccessException {
+  public void unShareDocument(String documentId, long destId) throws RepositoryException {
     documentFileStorage.unShareDocument(documentId, destId);
   }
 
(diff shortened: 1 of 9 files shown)


documents-services/src/main/java/org/exoplatform/documents/service/DocumentFileServiceImpl.java
```diff
@@ -314,14 +314,27 @@ public void updatePermissions(String documentId,  NodePermission nodePermissionE
         throw new IllegalStateException("Error updating sharing of document'" + documentId + " to identity " + destId, e);
       }
     });
+    if (nodePermissionEntity.getToUnShare() != null) {
+      nodePermissionEntity.getToUnShare().keySet().forEach(destId -> {
+        try {
+          unShareDocument(documentId, destId);
+        } catch (Exception e) {
+          throw new IllegalStateException("Error updating unSharing of document'" + documentId + " to identity " + destId, e);
```
ahamdi:
```suggestion
          throw new IllegalStateException("Error when unsharing the document {} with identity {}", documentId, destId, e);
```

exo-swf:

ahamdi (approved):

PR #1125, opened 2023-12-04
fix: apply border radius on the document application - EXO-67618

sofyenne:
Before this modification, no border radius was applied to the document application. This change will now implement the border radius on the document application


ahamdi (commented):
Could you add a message to describe the PR, even a simple one to exaplin what this is about

ahamdi (approved):

PR #1128, opened 2023-12-04
fix:  move drawer ko - EXO-67091

mkrout:
Prior to this change, if the user creates a shortcut to a folder in a destination that already contains a shortcut to the same folder. this breaks the tree view of the folder, since the service returns the full tree with all children and in this case, we will have an infinite loop.
This commit fixes the problem by retrieving only the first layer of children and then every time the user opens a folder in the tree, the subfolders will be retrieved from the server.

### DIFF
--- a/documents-storage-jcr/src/test/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorageTest.java
+++ b/documents-storage-jcr/src/test/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorageTest.java
@@ -764,6 +764,11 @@ public class JCRDocumentFileStorageTest {
     assertEquals(1, fullTreeItemList.size());
     assertEquals(3, fullTreeItemList.get(0).getChildren().size());
 
+    // withChildren is false, it should return list with just the parent folder children, sub folders should not have children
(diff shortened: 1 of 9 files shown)


documents-storage-jcr/src/test/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorageTest.java
```diff
@@ -691,7 +690,7 @@ public void testGetFullTreeData() throws Exception {
     when(sessionProvider.getSession("collaboration", manageableRepository)).thenReturn(session);
     when(identityManager.getIdentity(String.valueOf(ownerId))).thenReturn(ownerIdentity);
 
-    List<FullTreeItem> fullTreeItemList = jcrDocumentFileStorage.getFullTreeData(ownerId, folderId, identity);
+    List<FullTreeItem> fullTreeItemList = jcrDocumentFileStorage.getFullTreeData(ownerId, folderId, identity, true);
```
ahamdi, line 693: Could you add the case when the withChildren param is set to false ?

mkrout: Added

ahamdi (approved):

PR #1143, opened 2023-12-14
fix: No versioning when a documents is imported by a zip file - EXO-67504

mkrout:
Prior to this change, files imported from a zip file were not versioned.
This change creates a new version for all imported files.


azayati (approved):
LGTM

PR #1144, opened 2023-12-15
fix: No versioning when a documents is imported by a zip file - EXO-67504

Jihed525:
Prior to this change, files imported from a zip file were not versioned.
This change creates a new version for all imported files.
(cherry picked from commit c4980f91076b8571c2db02e20fa4c9039a8d39b0)


azayati (approved):
LGTM

PR #1147, opened 2023-12-18
fix: Fix wrong count for expiration delay when choosing day option - EXO-67306

IlhemEssaadi:
In this commit, we improve the accuracy of date calculations by introducing `parseInt(this.delayTypeTimes)` to parse the `delayTypeTimes` variable. This improvement enables us to calculate the new expiry date correctly.


azayati (approved):
LGTM

PR #1159, opened 2023-12-27
fix: Fix renaming documents with accents - EXO-67736

GouadriaHanen:
Before this change, when renaming a document by deleting the accent and making its name as the old name when it is cleaned by removing accents, an item already exists exception is thrown when trying to move srcPath to desPath since they are similar.
After this change, we will avoid to move srcPath to desPath when they are equal.


azayati (approved):
LGTM

PR #1175, opened 2024-01-23
fix: Change edit document redirection url from current portal name to meta portal name - EXO-69070

sofyenne:
Before this change, after opening the administration site and then accessing the document space application in new page to edit a file, we would be redirected to a wrong URL that displayed the administration page. This change is going to address the issue by modifying the edit document redirection URL from the current portal name to the meta portal name.


azayati (approved):
LGTM

PR #1240, opened 2024-04-17
fix: Document - folder view: Options z-index issue - EXO-70566

mkrout:
Prior to this change when user click on add item on dcocument up, the action  menu is hidden by the other components.
This fix put the button inside the menu template component to fix that


azayati (approved):
LGTM

PR #1242, opened 2024-04-18
fix: Document - folder view: Options z-index issue - EXO-70566 (#1240)

mkrout:
Prior to this change when user click on add item on dcocument up, the action menu is hidden by the other components.
This fix put the button inside the menu template component to fix that


akhanfir (approved):

Jihed525:
@mkrout I ebleive we don't need ton bcakcport fix as not reproduced on stable/6.5, wdyt ?

Jihed525 (approved):

PR #1243, opened 2024-04-18
fix: Documents : file saved on root path if Document app node name isn't default - EXO-71098

mkrout:
Prior to this, when document us is added to a space in page with name different from "Documents" and user ties to upload files under subfolders of the space, files are uploaded on the root folder
This issue is caused by wrong path computing that is based on default page name, the fix takes into account the current page name to get the correct folder path when uploading documents.


akhanfir (approved):

azayati (approved):
LGTM

PR #1320, opened 2024-12-23
feat: Implement fetching of trash element list - EXO-76054

sofyenne:


documents-services/src/main/java/org/exoplatform/documents/rest/DocumentFileRest.java
```diff
@@ -1298,5 +1302,58 @@ public Response importDocuments(@Parameter(description = "upload id", required =
...
+      @ApiResponse(responseCode = "401", description = "Unauthorized operation"),
+      @ApiResponse(responseCode = "500", description = "Internal server error"), })
+  public Response getDeletedDocuments(@QueryParam("sortField")
+                                      String sortField,
+                                      @Parameter(description = "Sort ascending or descending")
+                                      @QueryParam("sortDirection")
+                                      String sortDirection,
+                                      @Parameter(description = "Offset of results to return")
+                                      @QueryParam("offset")
+                                      int offset,
```
rdenarie, line 1325: can we have a default value for offset ? (zero for example)

documents-services/src/main/java/org/exoplatform/documents/rest/DocumentFileRest.java
```diff
@@ -1298,5 +1302,58 @@ public Response importDocuments(@Parameter(description = "upload id", required =
...
+      LOG.error("Invalid offset or limit parameters");
+      return Response.status(Response.Status.BAD_REQUEST).build();
+    }
+    TrashElementNodeFilter filter = new TrashElementNodeFilter();
+    if (StringUtils.isBlank(sortField)) {
+      sortField = "modifiedDate";
+    }
+    if (StringUtils.isBlank(sortDirection)) {
+      sortDirection = "desc";
+    }
```
rdenarie: Can we have default value for that ?

documents-services/src/main/java/org/exoplatform/documents/rest/DocumentFileRest.java
```diff
@@ -1298,5 +1302,58 @@ public Response importDocuments(@Parameter(description = "upload id", required =
...
+    filter.setOffset(offset);
+    filter.setLimit(limit);
+    filter.setSortField(getFromAlias(sortField));
+    try {
+      List<TrashElementNode> trashElementNodes = documentFileService.getDeletedDocuments(filter);
+      int size = documentFileService.countDeletedDocuments();
+      List<TrashElementEntity> trashElementEntities = trashElementNodes.stream()
+                                                                       .map(EntityBuilder::toTrashElement)
+                                                                       .collect(Collectors.toList());
+      CollectionEntity<TrashElementEntity> collectionEntity = new CollectionEntity<>(trashElementEntities,
```
rdenarie, line 1345: This is not performant, as you get all trash nodes, then you apply limit and offset AFTER. 
We need to pass offset and limit to documentFileService, which take care of offset and limit to return the correct list.

documents-services/src/main/java/org/exoplatform/documents/rest/model/TrashElementEntity.java
rdenarie, line 1: Is it possible to have only one class for trash object ? is it possible to merge TrashElementNode and TrashElementEntity ?

rdenarie, line 1: image

I do not understand why there is a renaming from documentsTrashList.vue to TrashElementEntity. 
Are you ok with this change ?

documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/TrashStorageImpl.java
```diff
@@ -47,25 +48,28 @@
...
+  private static final String    FILE_EXPLORER_PORTLET   = "FileExplorerPortlet";
 
-  private static final String TRASH_WORKSPACE = "trashWorkspace";
+  private static final String    TRASH_WORKSPACE         = "trashWorkspace";
 
-  private static final String TRASH_HOME_PATH = "trashHomeNodePath";
+  private static final String    TRASH_HOME_PATH         = "trashHomeNodePath";
 
-  private RepositoryService repositoryService;
+  private static final String    SELECT_FROM_TRASH_QUERY = "SELECT * FROM nt:base WHERE exo:restorePath IS NOT NULL";
```
rdenarie, line 62: offset and limit must be applied in this query

documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/TrashStorageImpl.java
```diff
@@ -328,11 +332,58 @@ public void removeDeadSymlinks(Node node) throws RepositoryException {
...
+  public int countDeletedDocuments() {
+    try {
+      StringBuilder sb = new StringBuilder(SELECT_FROM_TRASH_QUERY);
+      ManageableRepository manageableRepository = repositoryService.getCurrentRepository();
+      SessionProvider sessionProvider = sessionProviderService.getSystemSessionProvider(null);
+      Session session = sessionProvider.getSession(this.trashWorkspace, manageableRepository);
+      QueryManager queryManager = session.getWorkspace().getQueryManager();
+      Query query = queryManager.createQuery(sb.toString(), Query.SQL);
+      QueryResult queryResult = query.execute();
+      return (int) queryResult.getRows().getSize();
```
rdenarie: I'm afraid of performances of this query.
We know that, when you have millions of children for a node, the JCR have very poor performances.

I would prefer something like this : 
```suggestion
      NodeIterator iter = queryResult.getNodes();
      return iter.getSize();
```

rdenarie (changes requested):

documents-services/src/main/java/org/exoplatform/documents/rest/DocumentFileRest.java
sofyenne: done!

documents-services/src/main/java/org/exoplatform/documents/rest/DocumentFileRest.java
sofyenne: Limit and offset have already been passed to the service (included into the TrashElementNodeFilter) and applied to the query. This stream operation is only used to convert the returned list from `ElementTrashNode` to the DTO `ElementTrashEntity`!

documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/TrashStorageImpl.java
sofyenne: This is just a string used to refactor the code. It is used by multiple methods that do not use the offset and limit. In our case, the offset and limit are applied during the building of the query!

documents-services/src/main/java/org/exoplatform/documents/rest/DocumentFileRest.java
sofyenne: done!

documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/TrashStorageImpl.java
sofyenne: done!

documents-services/src/main/java/org/exoplatform/documents/rest/model/TrashElementEntity.java
sofyenne: It is not possible because TrashElementNode extends the abstract node, so there is no need to return unnecessary fields to the frontend!

sofyenne: I believe it is a GitHub mistake!

documents-services/src/main/java/org/exoplatform/documents/rest/DocumentFileRest.java
rdenarie: does both stream can be combined ?
Why do we use 2 ?

documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/TrashStorageImpl.java
rdenarie: :+1:

rdenarie (changes requested):

documents-services/src/main/java/org/exoplatform/documents/rest/DocumentFileRest.java
sofyenne: where is the second stream! , we have only one stream , the second one is a constructor of the CollectionEntity DTO ?!!

rdenarie: pff, sorry, I read to quickly ...

rdenarie (approved):

PR #1514, opened 2025-07-08
fix: Breadcrumb in doc app refers to trash and deleted shortcut instead of correct path of original folder - EXO-75899

akhanfir:
Before this change, when delete symlink from folder and open folder original then check breadcrumb, the breadcrumb of this folder appears trash. To fix this problem, when deleting the folder if it is symlink we delete the symlink and clear the symlinkNavHistory. After this change, the breadcrumb is displayed correctly.

### DIFF
--- a/documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/JCRDeleteFileStorageImpl.java
+++ b/documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/JCRDeleteFileStorageImpl.java
@@ -28,6 +28,7 @@ import javax.jcr.lock.LockException;
 import javax.jcr.version.VersionException;
 
 import org.apache.commons.lang3.StringUtils;
+import org.exoplatform.commons.utils.CommonsUtils;
 import org.exoplatform.documents.model.TrashElementNode;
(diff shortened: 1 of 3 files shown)


documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/JCRDeleteFileStorageImpl.java
```diff
@@ -164,7 +167,7 @@ public void deleteDocument(Session session,
           RequestLifeCycle.begin(container);
           try {
             documentsToDeleteQueue.remove(documentId);
-            moveToTrash(finalFolderPath, session, userIdentityId, favorite, checkToMoveToTrash);
+            moveToTrash(finalFolderPath, session, userIdentityId, favorite, isSymlink ? false : checkToMoveToTrash);
```
azayati:
```suggestion
            moveToTrash(finalFolderPath, session, userIdentityId, favorite, !isSymlink && checkToMoveToTrash);
```

azayati (changes requested):

azayati (approved):

PR #1525, opened 2025-07-10
fix: Copy link from 3dots - EXO-80141

mkrout:
This commit will fix the copy link to be relative to documents type and permissions.

### DIFF
--- a/documents-webapp/src/main/webapp/vue-app/documents/components/body/actions/CopyLinkMenuAction.vue
+++ b/documents-webapp/src/main/webapp/vue-app/documents/components/body/actions/CopyLinkMenuAction.vue
@@ -26,31 +26,28 @@ export default {
     publicAccessVisibilityChoice: 'COLLABORATORS_AND_PUBLIC_ACCESS'
   }),
   methods: {
+
     copyLink() {
-      let path = `${window.location.host}${eXo.env.portal.context}`;
-      if (eXo.env.portal.spaceId){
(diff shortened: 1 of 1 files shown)


documents-webapp/src/main/webapp/vue-app/documents/components/body/actions/CopyLinkMenuAction.vue
```diff
@@ -26,31 +26,28 @@ export default {
...
+        path =  `${window.location.host}${this.$documentsUtils.getEditorUrl(this.file,'view')}`;
       } else {
-        this.copy(path);
+        path = `${window.location.host}${this.$documentsUtils.getParentFolderUrl(this.file)}?documentPreviewId=${this.file.id}`;
       }
+      const input = document.createElement('input');
+      input.value = path;
+      document.body.appendChild(input);
+      input.select();
+      document.execCommand('copy');
```
boubaker: Just for my information, why not using `navigator.clipboard.writeText` knowing that [document.execCommand](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand) is deprecated ?

mkrout: yes, I just missed that, it was the case, we check if  'navigator.clipboard.writeText' is available for the current browser otherwise we use the execCommand I'll restore that. thanks

boubaker (approved):

PR #1561, opened 2025-07-21
fix: Copy link from 3dots - EXO-80141

mkrout:
This commit will fix the copy link to be relative to documents type and permissions.

### DIFF
--- a/documents-webapp/src/main/webapp/vue-app/documents/components/body/actions/CopyLinkMenuAction.vue
+++ b/documents-webapp/src/main/webapp/vue-app/documents/components/body/actions/CopyLinkMenuAction.vue
@@ -31,7 +31,7 @@ export default {
       this.loading = true;
       let path;
       if (this.file.folder){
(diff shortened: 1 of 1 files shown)


documents-webapp/src/main/webapp/vue-app/documents/components/body/actions/CopyLinkMenuAction.vue
```diff
@@ -30,7 +30,16 @@ export default {
     copyLink() {
       this.loading = true;
       let path;
-      if (this.isFileEditable())  {
+      if (this.file.folder){
+        path = `${window.location.host}${eXo.env.portal.context}`;
```
boubaker: It should may be user the whole `window.location.origin` including the protocol, WDYT ?

```suggestion
        path = `${window.location.origin}${eXo.env.portal.context}`;
```

boubaker (approved):

PR #1598, opened 2025-08-14
Merge exip 18

AzmiTouil:


documents-webapp/src/main/java/org/exoplatform/documents/portlet/DocumentGadgetPortlet.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  private String getOrCreateApplicationId(PortletPreferences preferences) {
+    String applicationId = preferences.getValue(APPLICATION_ID, null);
+    if (applicationId == null) {
+      applicationId = String.valueOf(random.nextLong() & Long.MAX_VALUE);
+      savePreference(APPLICATION_ID, applicationId);
+    }
+    return applicationId;
+  }
```
boubaker: As discussed, need to rework this portlet to rely on `Settings Name` of the portlet rather than redefining this mechanism.

```suggestion
```

boubaker (approved):

PR #1656, opened 2025-09-08
fix: Remove any dependencies to old doc preview - EXO-80993 (#1647)

mkrout:
This change will remove old documents preview and cleanup its dependencies, It will also and move search cards and ckeditor config from ecms


documents-webapp/src/main/webapp/WEB-INF/conf/documents/ckeditor-configuration.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+	This file is part of the Meeds project (https://meeds.io/).
+
+  Copyright (C) 2020 - 2024 Meeds Association contact@meeds.io
```
boubaker: Please replace the License Header to AGPL

documents-webapp/webpack.watch.js
```diff
@@ -5,7 +5,7 @@ const webpackProductionConfig = require('./webpack.prod.js');
 
 module.exports = merge(webpackProductionConfig, {
   output: {
-    path: '/exo-server/webapps/documents-portlet/',
+    path: 'D:\\\eXo\\servers\\platform-7.1.x-meeds-qaui-SNAPSHOT/webapps/documents-portlet/',
```
boubaker:
```suggestion
    path: '/exo-server/webapps/documents-portlet/',
```

boubaker (approved):

PR #1673, opened 2025-09-15
feat: UX to Open version - EXO-80872

mkrout:
This commit change the way documents versions are opened to be depending on document type.

### DIFF
--- a/documents-webapp/src/main/webapp/vue-app/documents/components/DocumentsMain.vue
+++ b/documents-webapp/src/main/webapp/vue-app/documents/components/DocumentsMain.vue
@@ -1613,17 +1613,21 @@ export default {
       return this.getFolderPath(this.folderPath);
     },
     showVersionPreview(version) {
-      return this.$attachmentService.getAttachmentById(version.originId)
+      return this.$attachmentService.getDocumentDetails(version.originId)
         .then(attachment => {
-          return this.$attachmentService.getAttachmentById(version.frozenId)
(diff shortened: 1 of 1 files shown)


documents-webapp/src/main/webapp/vue-app/documents/components/DocumentsMain.vue
```diff
@@ -1613,21 +1613,21 @@ export default {
...
     },
     showVersionPreview(version) {
-      return this.$attachmentService.getAttachmentById(version.originId)
+      return this.$attachmentService.getDocumentDetails(version.originId)
         .then(attachment => {
-          return this.$attachmentService.getAttachmentById(version.frozenId)
+          return this.$attachmentService.getDocumentDetails(version.frozenId)
             .then(file => {
-              file.downloadUrl =`${attachment.downloadUrl}?version=${version.versionNumber}`;
+              file.downloadUrl =`/rest/jcr/repository/collaboration${attachment.path}?version=${version.versionNumber}`;
```
boubaker, line 1620: Maybe use for the future URLs, the new WebDav URL, but not for now. In fact, this URL has to be deprecated and then deleted.

boubaker (approved):

documents-webapp/src/main/webapp/vue-app/documents/components/DocumentsMain.vue
mkrout: OK

PR #1724, opened 2025-10-16
fix: Transform DocumentPublic Access from handler to Portlet - EXO-81854

rdenarie:
Before this fix, documentPublicAccess use an handler. As we change the behaviour with meeds-io/MIPs#203 and switch to classic REST portlet, we need to update this portlet.


azayati (changes requested):
@rdenarie can you rebase ?

rdenarie:
> @rdenarie can you rebase ?

done @azayati

mkrout (approved):

azayati (approved):

PR #1817, opened 2025-12-26
feat: Display number of attachments in news activity - EXO-83426 - exoplatform/eXIP#33

akhanfir:


documents-webapp/src/main/webapp/vue-app/attachment/components/activity/ActivityAttachmentIcon.vue
```diff
@@ -52,7 +52,7 @@ export default {
       return this.attachments?.length > 9 ? '9+' : this.attachments?.length;
     },
     attachmentTitle() {
-      return this.attachmentsCount.toString().concat(' ').concat(this.attachments?.length > 1 ? this.$t('attachments.list') : this.$t('attachments.item'));
+      return this.attachments && this.attachments.length.toString().concat(' ').concat(this.attachments?.length > 1 ? this.$t('attachments.list') : this.$t('attachments.item'));
```
boubaker:
```suggestion
      if (!this.attachments?.length) {
        return null;
      }
      return  this.attachments.length === 1 ? `1 ${this.$t('attachments.item')}` : `${this.attachments.length} ${this.$t('attachments.list')}`;
```

boubaker (approved):

PR #1824, opened 2026-01-13
fix: Add placeholder page when the user accesses another user's personal drive -EXO-82012 

MayTekayaa:
Before this change, a user creates a document in his personal drive, then attaches it to an article in a space. Another user who is a member of the same space goes to the location of the first user's document, which is restricted because it is located in his personal drive. This change corrects this behavior and displays a placeholder page to the second user to indicate that he does not have access to this drive.


documents-webapp/src/main/webapp/WEB-INF/gatein-resources.xml
```diff
@@ -134,6 +134,35 @@
...
+      <depends>
+        <module>vue</module>
+      </depends>
+      <depends>
+        <module>vuetify</module>
+      </depends>
+      <depends>
+        <module>eXoVueI18n</module>
+      </depends>
+      <depends>
```
ahamdi: If you do not use Jquery, could you remove the dependency ?

documents-webapp/src/main/webapp/WEB-INF/jsp/restrictedDrive.jsp
```diff
@@ -0,0 +1,27 @@
...
+  * You should have received a copy of the GNU Affero General Public License
+  * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+%>
+<div class="VuetifyApp">
+  <div data-app="true"
+    class="v-application transparent v-application--is-ltr theme--light"
+    id="RestrictedDrive">
+    <script type="text/javascript">
+      require(['PORTLET/documents-portlet/RestrictedDrive'], app =>app.init('RestrictedDrive'));
```
ahamdi: The init function does not need a parameter

documents-webapp/src/main/webapp/vue-app/restricted-drive/main.js
```diff
@@ -0,0 +1,47 @@
...
+const url = `/documents-portlet/i18n/locale.portlet.Documents?lang=${lang}`;
+
+export function init() {
+  exoi18n.loadLanguageAsync(lang, url).then(i18n => {
+    // init Vue app when locale ressources are ready
+    Vue.createApp({
+      template: `<documents-restricted-drive id="${appId}" />`,
+      vuetify: Vue.prototype.vuetifyOptions,
+      i18n
+    }, `#${appId}`, 'Documents');
```
ahamdi: IMO you need to put the app name instead of Documents

ahamdi (changes requested):

rdenarie (approved):

ahamdi (approved):

PR #1827, opened 2026-01-26
fix: Fix trash restore path when parent folder is renamed - EXO-83976

hakermi:
Prior to this change, documents in the trash could not be restored if a parent folder was renamed or moved.

The restore mechanism relies on the `exo:restorePath` property, which becomes invalid when a parent folder path changes. As a result, restore operations failed because the stored path no longer matched the real structure.

This PR fixes the issue by creating a listener that tracks the move document event and update the restore path of trashed nodes accordingly to prevent the reference loss.

### DIFF
--- a/documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/TrashStorageImpl.java
+++ b/documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/TrashStorageImpl.java
@@ -48,8 +48,7 @@ import javax.jcr.query.QueryResult;
 import java.util.*;
 import java.util.stream.Collectors;
 
-import static org.exoplatform.documents.storage.jcr.util.NodeTypeConstants.EXO_DATE_MODIFIED;
-import static org.exoplatform.documents.storage.jcr.util.NodeTypeConstants.EXO_TITLE;
+import static org.exoplatform.documents.storage.jcr.util.NodeTypeConstants.*;
 
 public class TrashStorageImpl implements TrashStorage {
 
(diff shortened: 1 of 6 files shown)


documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/TrashStorageImpl.java
```diff
@@ -643,4 +642,28 @@ private void restoreFromTrash(String trashNodePath,
     restoreSession.save();
   }
 
+  @Override
+  public void updateRestorePath(String oldPath, String newPath) throws RepositoryException {
+    SessionProvider sessionProvider = sessionProviderService.getSystemSessionProvider(null);
+    Session session = sessionProvider.getSession(this.trashWorkspace, repositoryService.getCurrentRepository());
+
+    String queryStr = "SELECT * FROM nt:base WHERE exo:restorePath LIKE '" + oldPath.replace("'", "''") + "%'";
```
ahamdi, line 650: This query should be run just on the Trash folder to avoid looking in all documents. Besides we can run the query on nodes of type **exo:restoreLocation** instead of **nt:base** to limit the query to documents that have the property **exo:restorePath**

rdenarie:
@hakermi 
Can you complete your commit message ?

You clearly explain what is the problem, and why there is a problem, but not how you fixed it.

Thanks

hakermi:
> @hakermi Can you complete your commit message ?
> 
> You clearly explain what is the problem, and why there is a problem, but not how you fixed it.
> 
> Thanks

Done

documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/listener/MoveNodeListener.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @PostConstruct
+    public void init() {
+        listenerService.addListener("exo-document-moved", this);
+    }
+
+    @Override
+    public void onEvent(Event<String, String> event) throws Exception {
+        String oldPath = event.getSource();
```
ahamdi, line 46: This listener should run only if it is a parent node i.e a folder. There is no need to run this listener if the we move or rename documents (nt:file) for example

rdenarie (approved):

PR #1889, opened 2026-04-22
fix: Fix accessing a space drive using webdav - EXO-85437

MayTekayaa:
Prior to this change ,space drives could not be accessed via WebDAV. When users attempted to mount or browse a space drive, the server returned HTML error pages or directory listings that were incompatible with WebDAV clients.  
This fix will improve path handling so that space drives behave consistently with other drive types.

### DIFF
--- a/documents-webapp/src/main/webapp/vue-app/documents-user-setting/components/drawers/DocumentsWebdavMapDrivesDrawer.vue
+++ b/documents-webapp/src/main/webapp/vue-app/documents-user-setting/components/drawers/DocumentsWebdavMapDrivesDrawer.vue
@@ -265,7 +265,7 @@ export default {
       } else if (this.driveType === 'PERSONAL') {
         return `${window.location.origin}/webdav/drives/d/(${eXo.env.portal.userIdentityId})`;
       } else if (this.driveType === 'SPACE' && this.spaceIdentityId) {
-        return `${window.location.origin}/webdav/drives/d/(${this.spaceIdentityId})`;
+        return `${window.location.origin}/webdav/drives/d/${this.spaceIdentity.remoteId} (${this.spaceIdentityId})`;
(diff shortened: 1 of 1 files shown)


mkrout (approved):

documents-webapp/src/main/webapp/vue-app/documents-user-setting/components/drawers/DocumentsWebdavMapDrivesDrawer.vue
```diff
@@ -265,7 +265,7 @@ export default {
       } else if (this.driveType === 'PERSONAL') {
         return `${window.location.origin}/webdav/drives/d/(${eXo.env.portal.userIdentityId})`;
       } else if (this.driveType === 'SPACE' && this.spaceIdentityId) {
-        return `${window.location.origin}/webdav/drives/d/(${this.spaceIdentityId})`;
+        return `${window.location.origin}/webdav/drives/d/${this.spaceIdentity.remoteId} (${this.spaceIdentityId})`;
```
azayati, line 268: @MayTekayaa why  (${this.spaceIdentityId}) you keep this at the end of the url ?

azayati (changes requested):

azayati (approved):

documents-webapp/src/main/webapp/vue-app/documents-user-setting/components/drawers/DocumentsWebdavMapDrivesDrawer.vue
MayTekayaa: I kept (${this.spaceIdentityId}) in the URL because it identifies the space in WebDAV. The fix improves the handling of this path, so that spaces are accessible just like other drives.

PR #1893, opened 2026-04-24
Fix: Allow publisher without being redactor to create/ edit and delete document - EXO-85817

MayTekayaa:
Prior to this change, only redactors or managers in a space can manage documents. This change will expands permissions so that publishers, even if they are not redactors, can now create, edit and delete documents.

### DIFF
--- a/documents-api/src/main/java/org/exoplatform/documents/model/PermissionRole.java
+++ b/documents-api/src/main/java/org/exoplatform/documents/model/PermissionRole.java
@@ -18,5 +18,5 @@ package org.exoplatform.documents.model;
 
 public enum PermissionRole {
   ALL,
-  MANAGERS_REDACTORS;
+  MANAGERS_REDACTORS_PUBLISHERS;
 }
--- a/documents-services/src/main/java/org/exoplatform/documents/service/DocumentFileServiceImpl.java
+++ b/documents-services/src/main/java/org/exoplatform/documents/service/DocumentFileServiceImpl.java
@@ -533,8 +533,7 @@ public class DocumentFileServiceImpl implements DocumentFileService {
     Space space = spaceService.getSpaceById(spaceId);
     boolean canAdd = false;
     if (space != null) {
-      canAdd = !spaceService.hasRedactor(space) || spaceService.hasRedactor(space)
-          && (spaceService.isRedactor(space, currentUserName) || spaceService.isManager(space, currentUserName));
+      canAdd = spaceService.canRedactOnSpace(space, currentUserName);
     }
     return canAdd;
   }
--- a/documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/listener/RestrictContentCreationSpaceListener.java
+++ b/documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/listener/RestrictContentCreationSpaceListener.java
@@ -15,6 +15,8 @@ import org.exoplatform.social.core.space.model.Space;
 import org.exoplatform.social.core.space.spi.SpaceLifeCycleEvent;
 
 import javax.jcr.Node;
+import javax.jcr.NodeIterator;
+import javax.jcr.RepositoryException;
 import javax.jcr.Session;
 
 import java.util.ArrayList;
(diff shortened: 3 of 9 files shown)


documents-services/src/main/java/org/exoplatform/documents/service/DocumentFileServiceImpl.java
```diff
@@ -534,7 +534,8 @@ public boolean canAddDocument(String spaceId, String currentUserName) {
     boolean canAdd = false;
     if (space != null) {
       canAdd = !spaceService.hasRedactor(space) || spaceService.hasRedactor(space)
-          && (spaceService.isRedactor(space, currentUserName) || spaceService.isManager(space, currentUserName));
+          && (spaceService.isRedactor(space, currentUserName) || spaceService.isManager(space, currentUserName)
```
azayati: I think that you should use instead spaceService.canRedactOnSpace.
Otherwises, we have also other permissions to review not only can add document, we have edit/delete/upload and consult files/folders with restricted permissions, and the message 
Image should be updated to include space publishers

azayati (changes requested):

documents-services/src/main/java/org/exoplatform/documents/service/DocumentFileServiceImpl.java
MayTekayaa: done!
and for messages to update, it will be done using Crowdin

documents-webapp/src/main/webapp/vue-app/documents/components/body/table/DocumentActionMenu.vue
```diff
@@ -89,7 +89,6 @@ export default {
           iconExtension = this.$documentsIconsExtension?.[0]?.get?.(this.file?.mimeType);
         }
         extensions = extensions.filter(extension => extension.enabled(this.file, this.isMobile, this.currentView, this.isSearchResult, iconExtension));
-
```
azayati, line 92: revert this file

documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/util/JCRDocumentsUtil.java
```diff
@@ -598,6 +598,12 @@ public static void computeDocumentAcl(Node node, AbstractNode documentNode, Iden
       }
 
     }
+    for (MembershipEntry membership : aclIdentity.getMemberships()) {
+      if ("publisher".equalsIgnoreCase(membership.getMembershipType())) {
```
azayati: This will check all memberships even those of groups, you need to check only that the user is publisher in the space IMO no ?

azayati (changes requested):

documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/util/JCRDocumentsUtil.java
MayTekayaa: i have added a check if the current user is a publisher in the same space group as the document

documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/util/JCRDocumentsUtil.java
```diff
@@ -598,6 +602,10 @@ public static void computeDocumentAcl(Node node, AbstractNode documentNode, Iden
       }
 
     }
+    if (spaceGroup != null && aclIdentity.isMemberOf(new MembershipEntry(spaceGroup, "publisher"))) {
```
azayati: This should be done exactly as for "redactor" and "manager" cases.
See:

- https://github.com/exoplatform/documents/blob/aadddda3f0764ee4d12247615532067764b9fd78/documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/listener/RestrictContentCreationSpaceListener.java#L45
- https://github.com/exoplatform/documents/blob/aadddda3f0764ee4d12247615532067764b9fd78/documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/util/JCRDocumentsUtil.java#L605
- https://github.com/exoplatform/documents/blob/aadddda3f0764ee4d12247615532067764b9fd78/documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorage.java#L1783
- https://github.com/exoplatform/documents/blob/aadddda3f0764ee4d12247615532067764b9fd78/documents-services/src/main/java/org/exoplatform/documents/rest/util/EntityBuilder.java#L429

azayati (changes requested):

documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/util/JCRDocumentsUtil.java
MayTekayaa: These files are related to update permissions when changing access for a document (not related to adding  publisher or redactor) @azayati

azayati (changes requested):
to be reviewed as discussed

documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/util/JCRDocumentsUtil.java
azayati: to be reviewed as discussed @MayTekayaa

MayTekayaa:
> to be reviewed as discussed

Added as discussed @azayati
FYI, this change doesn't solve the problem for publishers in older spaces, as I explained to you.

azayati (dismissed):

documents-api/src/main/java/org/exoplatform/documents/model/PermissionRole.java
```diff
@@ -18,5 +18,5 @@
 
 public enum PermissionRole {
   ALL,
-  MANAGERS_REDACTORS;
+  MANAGERS_REDACTORS_PUBLISHERS;
```
boubaker, line 21: Add a new one and don't replace the existing behavior for retro compatibility

```suggestion
  MANAGERS_REDACTORS, MANAGERS_REDACTORS_PUBLISHERS;
```

documents-api/src/main/java/org/exoplatform/documents/model/PermissionRole.java
```diff
@@ -18,5 +18,5 @@
 
 public enum PermissionRole {
   ALL,
-  MANAGERS_REDACTORS;
+  MANAGERS_REDACTORS, MANAGERS_REDACTORS_PUBLISHERS;
```
boubaker: Now that the upgrade plugin is added, we can delete this. sorry, I didn't saw the Data Upgrade PR before


```suggestion
  MANAGERS_REDACTORS_PUBLISHERS;
```

boubaker:
Need to apply this recursively IMO when switching from redactional space to non restricted space and vice versa

documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/listener/RestrictContentCreationSpaceListener.java
```diff
@@ -104,4 +89,40 @@ private void changePermissionsForSpaceMembers(Space space, boolean readOnlyForMe
       }
     }
   }
+
+  private void applyPermissionsRecursively(Node parentNode, Space space, boolean readOnlyForMembers) throws RepositoryException {
+    NodeIterator children = parentNode.getNodes();
+    while (children.hasNext()) {
+      Node child = children.nextNode();
+      if (child.isNodeType("exo:privilegeable")) {
```
boubaker: Apply the new ACL only when the permissions fit the old Space Redactional Flag (change to redactors only when open to all & vice versa)

AzmiTouil: Done!

PR #1917, opened 2026-06-08
fix: Scope lastView localStorage key per space to prevent view preference conflicts between spaces - EXO-83876

AzmiTouil:


documents-webapp/src/main/webapp/vue-app/documents/components/DocumentsMain.vue
```diff
@@ -508,7 +508,8 @@ export default {
       }
     },
     initSettings() {
-      const lastView = eXo.env.portal.spaceId ? localStorage.getItem('lastView') : 'folder';
+      const lastViewStorageKey = eXo.env.portal.spaceId ? `lastView-${eXo.env.portal.spaceId}` : 'lastView';
```
boubaker, line 511: May be adding the `userName` too in case, the same browser is used by multiple users, but this still optional I let you choose.

```suggestion
      const lastViewStorageKey = eXo.env.portal.spaceId ? `lastView-${eXo.env.portal.userName}-${eXo.env.portal.spaceId}` : `lastView-${eXo.env.portal.userName}`;
```

boubaker (approved):
Minor option to include if possible, else no worry it's ok for me

PR #1919, opened 2026-06-08
fix: Allow non-admin users to access WebDAV subfolders in spaces - EXO-86876

AzmiTouil:
Use system session for path existence checks in checkResourceExists to avoid permission failures for non-admin users


documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/webdav/plugin/WebdavReadCommandHandler.java
```diff
@@ -791,25 +797,30 @@ private String getDisplayName(long identityId) {
   }
 
   private String checkResourceExists(Session session, String path) throws RepositoryException, WebDavException {
+    Session systemSession = getSystemSession();
```
boubaker, line 800: Need to discuss more about this with @srenault-meeds

PR #1932, opened 2026-06-10
fix: Fix Folder/File management when content restriction is ON - EXO-86607

AzmiTouil:
Fix feedbacks


documents-services/src/main/java/org/exoplatform/documents/service/DocumentFileServiceImpl.java
```diff
@@ -534,7 +534,7 @@ public boolean canAddDocument(String spaceId, String currentUserName) {
     boolean canAdd = false;
     if (space != null) {
       canAdd = !spaceService.hasRedactor(space) || spaceService.hasRedactor(space)
```
boubaker, line 536: User `SpaceService.canRedactOnSpace` as discussed and perhaps continue the fix made on https://github.com/exoplatform/documents/pull/1893 to apply the required permission changes recursively on Space Folder.

PR #1941, opened 2026-06-15
fix: Fix the XSS vulnerability in SVG files - EXO-86346

MayTekayaa:
This change will fix the XSS vulnerability in SVG files.


boubaker (commented):
Shouldn't we do the same in all Vue files where we upload images and files ?

PR #1942, opened 2026-06-15
fix: Allow publisher without being redactor to create/edit/delete documents and apply permissions recursively with state check - EXO-85817

AzmiTouil:


documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/listener/RestrictContentCreationSpaceListener.java
```diff
@@ -104,4 +126,66 @@ private void changePermissionsForSpaceMembers(Space space, boolean readOnlyForMe
...
+      NodeIterator children = parentNode.getNodes();
+      while (children.hasNext()) {
+        if (cancelFlag != null && cancelFlag.get()) {
+          LOG.info("Permission restriction cancelled for space {}, stopping recursion",
+                   space.getPrettyName());
+          return;
+        }
+        Node child = children.nextNode();
+        try {
+          if (child.isNodeType("exo:privilegeable")) {
```
boubaker: Add further check whether the Node has custom permissions or not. Apply defaut permissions only on non-custom permission based nodes.

documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/listener/RestrictContentCreationSpaceListener.java
```diff
@@ -104,4 +126,66 @@ private void changePermissionsForSpaceMembers(Space space, boolean readOnlyForMe
...
+                                           boolean readOnlyForMembers,
+                                           AtomicBoolean cancelFlag) {
+    try {
+      NodeIterator children = parentNode.getNodes();
+      while (children.hasNext()) {
+        if (cancelFlag != null && cancelFlag.get()) {
+          LOG.info("Permission restriction cancelled for space {}, stopping recursion",
+                   space.getPrettyName());
+          return;
+        }
```
boubaker: We should manage this using a custom checked exception which has to be catched deeply in the Recursive Call Tree in order to stop the execution right away if the Space is switched On/Off in the meantime.

boubaker (dismissed):

boubaker (approved):

PR #1968, opened 2026-06-25
fix: include symlink target paths in ES search queries - EXO-67331

AzmiTouil:
This change will add symlink target path resolution to Elasticsearch document search queries.


documents-storage-jcr/src/main/java/org/exoplatform/documents/storage/jcr/JCRDocumentFileStorage.java
```diff
@@ -426,6 +432,41 @@ public List<FileNode> search(DocumentTimelineFilter filter,
     }
   }
 
+  private List<String> findSymlinkTargetPaths(Session session, String path) throws RepositoryException {
+    List<String> targetPaths = new ArrayList<>();
+    String query = "SELECT * FROM nt:base WHERE jcr:path LIKE '" + path + "/%' AND jcr:primaryType = 'exo:symlink'";
```
boubaker:
```suggestion
    String query = "SELECT * FROM exo:symlink WHERE jcr:path LIKE '" + path + "/%'";
```

boubaker (approved):

